PR #57, opened 2014-12-30
Add Bouncy Castle - Cryptographic C# API

juancri:


quozd:
Hi @juancri 
Thank you for your contribution. I can not merge this PR, because it doesn't comply with [contribution guidelines and quality standard](https://github.com/quozd/awesome-dotnet/blob/master/CONTRIBUTING.md). Bouncy Castle is too outdated and not maintained any more.

juancri:
Sadly, this library implements a lot of useful algorithms. Do you think a section for unsupported libraries could be added? It may include a big warning.

0xdeafcafe:
I agree with @juancri, having a section that comes with a warning is a fantastic idea.

quozd:
Yes, you're right. But we need to define some strict rules and guidelines for this section first.

quozd:
The main idea is to keep only extremely useful but unsupported projects in the list. I don't want to have here crap which somebody used someday and there weren't any updates for a 2+ years.

0xdeafcafe:
Could I suggest a second markdown file linked to in the primary of old and unsupported libraries?

quozd:
Sounds good for me :+1:

juancri:
By the way, take a look at the bouncy castle git repository:
https://github.com/bcgit/bc-csharp

It seems to be updated regularly and maintained.

quozd:
:open_mouth: This looks interesting.
OK. Let's add link to git repo instead of website, which looks dead. It will be much less confusing for people.

@juancri Update PR and I will merge it

quozd:
@juancri will you update your PR so I can merge it?

quozd:
Close because of inactivity

PR #61, opened 2015-01-27
Added Griffin.Framework

jgauffin:


quozd:
@jgauffin Griffin looks promising, but it is still too young to say that it is awesome project and useful for community. Right now it doesn't comply with our quality standards. Come back later when Griffin will be more popular.

PR #69, opened 2015-02-27
add some libraries from mono

plantain-00:


vladen:
@plantain-00, since you're introducing the "Compression" section, it's reasonable to migrate "DotNetZip" and "SharpZipLib" from Misc.

vladen:
The "XSP" project can be placed under already existing "Web Servers" section, which reflects the nature of the "XSP" project more precisely (from my perspective).

plantain-00:
@vladen , Thank you, I am glad that someone reviews it.

vladen:
My excuses for delay with reviewing ;)

PR #70, opened 2015-03-01
Update README.md

romanov:
Added MaterialSkin


vladen:
I'd like to notice the fact that project is not mature enough yet. Lack of documentation, poor tab order support, incomplete set of controls. However, the idea is nice. I think we should give the project some more time. What is your opinion, @quozd?

quozd:
Agree,
@romanov feel free to come back when project become more mature

PR #71, opened 2015-03-19
add msbuild

plantain-00:


quozd:
:+1: But I think Build Automation category will be much better. What do you think?
cast @vladen

plantain-00:
OK

quozd:
Good job! Thank you

PR #76, opened 2015-04-17
Add Fody and move Mono.Cecil to Reflection category

VPashkov:


gep13:
Great additions! /cc @simoncropp :+1:

quozd:
Good job :+1:

SimonCropp:
Not sure fody should be in reflection

quozd:
@SimonCropp feel free to propose better section :) We're alway open for contributions.

VPashkov:
@SimonCropp may be AOP section will suit better?

SimonCropp:
@VPashkov  i think "Assembly manipulation" is better since then that covers Cecil and Fody. Also technically, while you can achieve AOP with fody, it is not just AOP. For example this https://github.com/Fody/Caseless has no "apsect".

VPashkov:
@SimonCropp Agreed. Thanks.

quozd:
Good. Thanks guys.
@VPashkov can you please move Fody and Cecil to the new section. Then I will merge.

PR #77, opened 2015-04-22
Rename [Reflection] section to [Assembly Manipulation]

VPashkov:
It seems that [Assembly Manipulation] name better describes Fody and Mono.Cecil than [Reflection]


quozd:
Thank you

PR #78, opened 2015-04-24
Hello world F#

haf:
- Adding Suave under API/Frameworks
 - Adding Logibit Hawk under Authentication
 - Adding F#, F* under Compiler
 - Changing DotNetZip to its new home on my github; the codeplex site is dormant
 - Adding BouncyCastle under Crypto
 - Adding Http.fs under HTTP
 - Adding Logary under Logging
 - Some whitespace changes
 - Removing OpenWrap as it's dormant/dead since at least 2012
 - Adding Chiron to Serialization
 - Adding DotLiquid to TemplateEngine
 - Adding Suave to web servers, websockets (or Server-Sent Events in this case)

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,6 @@ Thanks to all [contributors](https://github.com/quozd/awesome-dotnet/graphs/cont
 ## Compiler
 
 * [F#](https://github.com/fsharp/fsharp/) -  The F# compiler, core library and tools - a functional programming language for safer, faster, better code writing.
-* [F*](https://github.com/nikswamy/FStar) - A dependently- and session-typed extension of the F# compiler that lets you prove your programs to great accuracy.
 * [Roslyn](https://github.com/dotnet/roslyn) - The .NET Compiler Platform ("Roslyn") provides open-source C# and Visual Basic compilers with rich code analysis APIs. It enables building code analysis tools with the same APIs that are used by Visual Studio.
 * [VisualFSharp](https://github.com/Microsoft/visualfsharp) - The Visual F# compiler and tools
 * [Mono-basic](https://github.com/mono/mono-basic) - Visual Basic Compiler and Runtime
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ Thanks to all [contributors](https://github.com/quozd/awesome-dotnet/graphs/cont
 ## API
 
 * Frameworks
+  * [Suave.IO](http://suave.io/) - Framework/library/web server that makes you cry tears of joy after finishing your project ahead-of-time when you look at the beautiful code you've written in F#.
   * [ASP.NET WebAPI](http://www.asp.net/web-api/) - Framework that makes it easy to build HTTP services that reach a broad range of clients, including browsers and mobile devices
   * [ServiceStack](https://github.com/ServiceStack/ServiceStack) - Thoughtfully architected, obscenely fast, thoroughly enjoyable web services for all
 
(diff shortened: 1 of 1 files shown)


tpetricek:
I think F* might be a bit of a stretch at the moment - it is more of a research project than a directly useful .NET technology. But it is certainly interesting thing too!

All the other projects are fantastic, well-tested and used by the community, so +1 for all of them!

haf:
Hehe, alright. I can remove that.

quozd:
Hi guys. Thank you for contribution. Great job. But I have to review all this links before merging and it will take some time.

vladen:
@quozd, does it make sense to sort each category alphabetically to be impartial and try to avoid repetitions to be concise?

quozd:
@vladen right. sorting is essential. uniqueness should be a priority, but not mandatory as some projects do too much stuff.

I took a look on proposed projects and not sure that we'll merge all of them (some are too outdated), but I don't have enough time right now to make the investigation :(

vladen:
Chiron seems to be not mature enough, but interesting though. Maybe there's a reason to inroduce very special category for such promising projects?

quozd:
I think we need to add some kind of labels for F# projects. F# is great and these projects are very promising, but they are not as mature as C# projects in the list. And I think they can't be, F# is not so popular.

What do you think, guys?

vladen:
Maybe it's time to split into two streams? Say F# itself is pretty adult and some its APIs are ready to use from any .NET compatible language. But some other APIs are too F# specific and cannot be considered as widely applicable. Let me also remind, this is awesome-dotnet repo, not awesome-F#.

vladen:
However, my confident vote for DotLiquid, logary and fsharp (just because it's nothing less than global initiative). Albacore requires special treatment, dealing with .Net and mono but being written on Ruby. Suave is awesome but IMHO too fsharp specific. Is it possible to use it from any other language, does it make sense? The same to Logibit Hawk. Not sure, I'm right here.

tpetricek:
I have not used all the projects that have been mentioned here (but they all look "awesome" to me and they all are relevant to "dotnet", which makes me think that the PR is quite appropriate!)

Adding a tag to the projects that are more focused on F# sounds good, but I would not make this a top-level category (having a common .NET platform means that we do not need "awesome-csharp" and "awesome-fsharp" :-)). If we wanted to be strict about language tagging, it would mean that libraries that use some constructs which make them not-very-nice-to-use from F# would need to have "C#-specific" tag...

I have first-hand experience with DotLiquid, Logary, Http.fs and they are all great. The other projects in the PR look very useful to me, but I have not actually used them.

Also, I think Suave should *definitely* be included, now that [even Scott Hanselman is using it](http://www.hanselman.com/blog/RunningSuaveioAndFWithFAKEInAzureWebAppsWithGitAndTheDeployButton.aspx) :-). On a more serious note, it is a great way to run simple .NET servers on Azure, and (through the perspective of "micro-services") can be easily combined with C# projects!

quozd:
OK, let's do the following:

1. Define the most valid category for Suave. I think [Web Frameworks](https://github.com/quozd/awesome-dotnet#web-frameworks) is the best choice.
2. Remove DotNetZip from this PR as it will be added in scope of #84 
3. If the project is very F# specific or has a much better alternative in C#, mark it with `[F#]` tag in the description. F# compiler shouldn't be marked :)

Thanks

tpetricek:
@quozd  +1, I think this is a good plan!

vladen:
@quozd +1

haf:
Ok, sure, do you need me to rebase, or can you merge in what you want?

quozd:
@haf  It will be great if you do these changes :)

quozd:
@haf will you update your PR so I can merge it?

haf:
I'm too short on time right now. Feel free to edit my PR's content.

vladen:
Merged all but Chiron (too young).

PR #81, opened 2015-05-01
Add OpenVR

plantain-00:


quozd:
@plantain-00 how is it related to .NET? I can see some binaries (dll and so) and cpp examples. Could you please describe why should we add this repo to our list. Thanks

plantain-00:
Sorry about that, I didn't check it well.
I saw 51.6% of them is C#, but didn't realize that it doesn't contain any code yet.

PR #82, opened 2015-05-02
Adding Spring.Net as Application Framework too

luizcarlosfaria:
On my first contribution, Application Framework category was didn't exist. Now this framework can be part of this category too.


quozd:
Thank you

PR #90, opened 2015-05-20
 Language-Ext

louthy:
Adding Language-Ext
A popular library (nearly 600 followers) for abusing C# to make it easier to write in a functional style:
https://github.com/louthy/language-ext


quozd:
:+1: thank you

quozd:
Sorry, @louthy I think API is not right category for this. I reverted PR. Could you please add your link to Misc?

louthy:
Hi @quozd - yep, I can do that.  I was a bit unsure of the best place for it.  I wonder, before I resubmit under Misc whether it'd be better to have a new section called 'Functional' or 'Functional Extensions' or something like that?  There are quite a few useful .NET libraries that are general case functional libraries which don't fit into the categories.  It seems a shame to lump them into a non-descriptive Misc section.

Your call obviously, happy to go with whatever you advise.

Paul

PR #94, opened 2015-05-26
Update README.md

hlesesne:


quozd:
It's not opensource, but let it be. Thank you

PR #96, opened 2015-05-26
Added MyGet

xavierdecoster:
cc @MyGet

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ Thanks to all [contributors](https://github.com/quozd/awesome-dotnet/graphs/cont
 ## Continuous Integration
 * [TeamCity](http://www.jetbrains.com/teamcity/) - Ready to work, extensible and developer-friendly build server — out of the box
 * [CruiseControl.NET](http://www.cruisecontrolnet.org/) - an Automated Continuous Integration server, implemented using the .NET Framework
-* [MyGet](http://www.myget.org/) - Continuous Integration and Deployment, Hosted Package Repository for NuGet, NPM, Bower and VSIX (Free for OSS).
+* [MyGet](http://www.myget.org/) - Continuous Integration and Deployment, Hosted Package Repository for NuGet, NPM, Bower and VSIX. **[Free for OSS]**
 
 ## Cryptography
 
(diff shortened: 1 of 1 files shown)


quozd:
Hi @xavierdecoster , thank you for your contribution.

@vladen what do you think about this project? It's free for OSS, but I don't want to make our list full of ads.

vladen:
Hmm. Good question. There was a proposal to split into two streams, one for OSS and another for awesome but proprietary or partially/fully commercial solutions. From the awesomeness perspective we should enlist as much as possible software that can be really useful. Does commercial nature of a product break its awesomeness? What about those developers who are ready to pay for cool and useful instrument? Imho, Teamcity is great and have to be known. The same for myget. Not so far ago I've got a question about free private NuGet repo, where to get it. Here is the answer.

vladen:
It was wrong button )

ghuntley:
Just because a product costs money does not mean that it is not awesome.
Developers should be prepared to pay for tools that provide value. Having a
commercial ecosystem is one of the indicators that companies use to gauge
maturity level.

quozd:
I'm not saying that commercial software is not awesome. I just want to keep this list free from advertisements. 

To add any proprietary software we should define some strict guidelines which should help up decide whether some software can be added or not.  If we start adding proprietary projects based on our own opinion and feelings then guys will start asking me "why this one is in the list, but my project was rejected".

vladen:
I think at least another list should be allocated for ads-like, proprietary or dual licensed software to keep clean the current one. Imho the main criteria for inclusion - usufulness for community. Other applicable criterias are maturity, stability, high value and perhaps presence of special licensing schema for open-source projects. Let's brainstorm this onwards.

quozd:
@xavierdecoster  I've added contribution guidelines for a proprietary software. Please, change your PR to comply with these requirements before we can merge it.

xavierdecoster:
Just added the Free-For-OSS tag. Note that MyGet has a Freemium model so anyone can join the Free offering. Those requesting the OSS plan get an extra special offering.

quozd:
@xavierdecoster MyGet link should hav both [$] and [Free for OSS] tags. Please, fix. Then i will merge

xavierdecoster:
Added. Thanks!

quozd:
:+1:

PR #97, opened 2015-05-26
Update README.md

Biocodr:
Added the TomP2P.NET library, feel free to move into another category. Thank you!


quozd:
Hello @Biocodr , thank you for your contribution.

Unfortunately, we can't accept this PR right now because project is too young. It lacks of documentation and is not yet ready.

PR #99, opened 2015-05-28
Added 2 template engine libs - Nustache and DotLiquid

Jacek-Pintera:
Hi guys,
I'm adding two nice libs for processing templates in .NET -  Nustache and DotLiquid


quozd:
Thank you

PR #115, opened 2015-06-20
Add Spring.NET RestTemplate to HTTP

canton7:


quozd:
Hello, thank you for your contribution. Unfortunately, this project looks abandoned. No commits for a long time (https://github.com/spring-projects/spring-net-rest). This violates our quality standards. I have to decline this PR.

canton7:
Fair enough - the quality standards do exclude established projects that have fixed their issues, in favour of newer, less stable projects with more issues :wink: I'd argue that lack of activity itself shouldn't be a barrier: ignored open issues (that are reasonable) should be a bad marker, for sure, but no issues and no activity is IMO fine.

canton7:
Actually re-reading `CONTRIBUTING.md` I see that's untrue. Is the one open issue, with no response, the thing you're picking up on? 

I'm not particularly invested in getting this merged: I used to use it, but now find it a bit ugly and cumbersome. Just trying to understand what exactly your criteria are :smile:

quozd:
We don't exclude established projects, but we don't like unsupported projects. I want to be sure, that anyone who will work with software from this list will be able to get good support and bug fixes. No repo activity means no bug fixes and support. I don't believe that Spring REST is bug free. If there is some other place where users can get solution for theirs issues - just let me know and I will check it.

sudo cast @vladen :)

canton7:
That's fair, thanks for responding.

PR #116, opened 2015-06-20
Adding DotImaging framework

dajuric:
The framework sets focus on .NET native array as primary imaging object, offers extensibility support via extensions, and provides unified platform-abstract imaging IO API. The roots of the framework can be found in https://github.com/dajuric/accord-net-extensions which is already on the list.


quozd:
Hey @dajuric. Thank you for your contribution, but I think this project is too young, there is no tests. /cc @vladen

dajuric:
OK, could you then put that pull request to a waiting list, so when the project gains some popularity you can add it ?

quozd:
We don't have a special waiting list. But you're welcome to create new PR as soon as your project will be mature enough. Then I will proudly merge it :) Thanks

PR #120, opened 2015-07-10
Added DNTProfiler.

VahidN:
Added DNTProfiler. It's an EF 6.x and NH 4.x profiler.


quozd:
Hi, thank you for your contribution. What about project documentation?

@vladen Im not sure about this one, what do you think? Looks very interesting, but too young as well.

vladen:
Young but useful. @VahidN, it would be worth to make more docs (maybe video guide as well to cut the learning curve). Thing seems to be very helpful for those, who work with EF, NH.

PR #121, opened 2015-07-10
Add DynamicImage

tgjones:


quozd:
Hi, thanks for your contribution. Looks like this project is not so actively maintained. Lots of old issues, more than half a year without commits. @tgjones, Is it alive?

@vladen what do you think?

tgjones:
@quozd Almost all of the open issues are actually features. There's only one actual bug, which I haven't yet been able to resolve, but fortunately it doesn't affect many people.

There's little activity because the project is stable now - it's been around for a few years, and is currently on v3.1.1.

quozd:
@tgjones thanks for clarification. Merged

tgjones:
Great, thanks!

vladen:
Cool stuff, with no doubts!

PR #122, opened 2015-07-10
Update README.md

wallisb:
Added Warewolf ESB under Queue category


quozd:
Looks interesting, thanks

PR #125, opened 2015-07-13
Add FluentValidation library.

VPashkov:
A small validation library for .NET that uses a fluent interface and lambda expressions for building validation rules.


quozd:
Thanks, good job.

PR #126, opened 2015-07-13
KVM -> DNVM name change

ktos:
Name of KVM tool for ASP.NET 5 was changed to DNVM some time ago, let's reflect this change here. Also added small tool description from project page.


quozd:
Great, thanks. Missed that somehow :)

PR #127, opened 2015-07-15
Add Shouldly - Should testing for .net

VPashkov:
Shouldly is an assertion framework which focuses on giving great error messages when the assertion fails while being simple and terse.


VPashkov:
I don't know why another commit is

PR #131, opened 2015-07-24
Add Reactive Extensions (Rx.Net) library

VPashkov:
The Reactive Extensions (Rx) is a library for composing asynchronous and
event-based programs using observable sequences and LINQ-style query operators.


quozd:
Nice, dude. Thank you

PR #133, opened 2015-08-03
Include link to MetadataExtractor project

drewnoakes:


devkhan:
You have created a very good project, but I think we should give it some more time, let's see where this project goes. Plus, there is lack of docs. I'll keep the PR open for some time. 

Good luck with the project. :+1:

drewnoakes:
Thanks for the feedback. Actually the project has been going strong since 2002 as a [Java library](https://github.com/drewnoakes/metadata-extractor) which has a thorough set of docs in its [wiki](https://github.com/drewnoakes/metadata-extractor/wiki). I'm in the process of unifying the documentation into a single place across Java/.NET APIs, and updating the project's home page on my own web site.

For reference, the Java version is downloaded from Maven Central over 50,000 times a month, recently and is on the rise (that peak is 134,325 downloads, and I have no idea why!):

image

The .NET implementation has the same capabilities as the Java version. The APIs are almost identical (save for use of .NET idioms around naming, use of structs, `TryGet*` and so forth).

Given the capability and popularity of the Java version, I am working to promote the .NET implementation amongst the .NET community, hence this PR.

devkhan:
Okay, since this is a project based on a successful one, I'll merge to promote it.

drewnoakes:
Thanks!

PR #135, opened 2015-08-05
Added Hangfire

jenyayel:


quozd:
Hey, we already have it in Queue category

jenyayel:
Great then! It was strange to me that you don't have it.

PR #136, opened 2015-08-05
Add Paradox Game Engine

VPashkov:


devkhan:
@quozd Should I add this one. This looks promising. But its still a new project and in beta stage too.

quozd:
Well, this looks very interesting, has good documentation and forums. Quite good though young. Will merge

PR #162, opened 2015-09-09
Add SmartStoreNET

VPashkov:


devkhan:
You already have provided this in the trading category. I think may be we shouldn't repeat this.

VPashkov:
Oops. 
My last PR was incorrect. I already sent another PR to fix it.
The correct name for project in Trading category is StockSharp.
This PR adds an actual SmartStoreNET project.

PR #171, opened 2015-09-21
Adds Wyam

daveaglick:
Wyam is a static content generator and can be used to generate web sites, produce documentation, create ebooks, and much more. Everything is configured by chaining together flexible modules (that can be easily written and extended).


quozd:
I think we can create a new category for ssg.

daveaglick:
Great - I'll push an update to include [Sandra](https://github.com/Sandra/Sandra.Snow) and [Pretzel](https://github.com/Code52/pretzel) as well.

daveaglick:
Added a new Static Site Generators section

PR #179, opened 2015-09-29
Add DiffSharp

gbaydin:


devkhan:
This seems pretty new. Let it grow for sometime first.

PR #186, opened 2015-10-05
Added TLSharp client for Telegram API

sochix:


devkhan:
Still in early phase, needs more time.

PR #194, opened 2015-10-12
Add Handlebars.Net

StevenDevooght:


devkhan:
Still in early stage. Many features are not implemented yet. Needs more time.

PR #199, opened 2015-10-23
add Micro Framework category and NETMF interpreter

ktos:
Hey,
I'd love to add some .NET Micro Framework projects to at least show it exists and lives (again!), starting with [.NET Micro Framework itself](https://github.com/NETMF/netmf-interpreter).

If you don't know - NETMF is an IL interpreter from Microsoft for small embedded devices (let's say - Arduino class), and version 4.4 was just released (and it's on GH since a few months ago).

I've prepared a new category "Micro Framework" with a link to NETMF to it, is it ok?

The problem is, however, that many of libraries for NETMF aren't actively maintained anymore, so they won't pass through guidelines probably...

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ Thanks to all [contributors](https://github.com/quozd/awesome-dotnet/graphs/cont
   * [Mathematics](#mathematics)
   * [Media](#media)
   * [Metrics](#metrics)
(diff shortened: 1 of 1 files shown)


quozd:
That's ok. I think there are at least some good libraries for MF. Thank you

PR #200, opened 2015-10-25
Add link to the Excel-DNA project (add-ins for Excel using .NET)

caioproiete:
Excel-DNA is an independent project to integrate .NET into Excel. With Excel-DNA you can make native (.xll) add-ins for Excel using C#, Visual Basic.NET or F#, providing high-performance user-defined functions (UDFs), custom ribbon interfaces and more. Your entire add-in can be packed into a single .xll file requiring no installation or registration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -456,6 +456,7 @@ metadata in media files, including video, audio, and photo formats
 
 ## Office
 
+* [ExcelDna](https://github.com/Excel-DNA/ExcelDna) - ExcelDna makes it easier to create and deploy Excel Add-Ins using C#, F# or VB .NET
 * [ClosedXML](https://closedxml.codeplex.com/) - ClosedXML makes it easier for developers to create Excel 2007/2010 files
 * [NPOI](http://npoi.codeplex.com/) - This project is the .NET version of POI Java project at http://poi.apache.org/.
 * [EPPlus](http://epplus.codeplex.com/) - EPPlus is a .net library that reads and writes Excel 2007/2010 files using the Open Office Xml format (xlsx).


quozd:
Hello, thank you for your contribution. Unfortunately this project violates our quality standards. Please, come back later with good test suite and we'll merge it.

caioproiete:
Hi @quozd,

I think you need to have some flexibility on the `Tests` requirement for projects like these. Excel-DNA runs inside Excel and calls directly on Excel APIs and as such, it will never have a good coverage in terms of automated unit tests (it is just not possible), and integration tests are quite tricky given that [Microsoft doesn't support running Excel on a server](https://support.microsoft.com/en-us/kb/257757), so we'll likely never see an AppVeyor build server (or similar) that could run such a suite of integration tests that require Excel to work.

To compensate for that, the core developer is very diligent, and goes through a long stabilization process before shipping a vNext, releasing lots of beta and release candidate versions during this process.

If the above is not enough to convince you, I don't know what is :)

quozd:
Hi @caioproiete. We're always open to discuss any of the proposed projects and want to add the best software to our list. Unfortunately sometimes we're not able to have a deep dive into the project and have to decide according to some formalized requirements only.

I believe that project is useful but I'm not sure if it is stable enough, I just can't verify it. 

@vladen, @devkhan what do you think about this one?

quozd:
cc @Jeffiy

Jeffiy:
LGTM, it seems not easy to be tested, we can give a chance too.

quozd:
OK, lets merge it then

caioproiete:
Perfect. Thanks

PR #202, opened 2015-10-25
Add DbUp - library for deployment of changes to SQL Server databases

caioproiete:
DbUp is a .NET library that helps you to deploy changes to SQL Server databases. It tracks which SQL scripts have been run already, and runs the change scripts that are needed to get your database up to date. http://dbup.github.io


gep13:
@JakeGinnivan FYI...

PR #212, opened 2015-11-03
Add Jexus

Jeffiy:


Jeffiy:
The Jexus is very awesome and it has been used in production, you can read more about it at https://jexus.codeplex.com and https://en.wikipedia.org/wiki/Jexus
cc @quozd @devkhan @vladen

PR #213, opened 2015-11-03
Add SuperSocket

Jeffiy:
Add SuperSocket and alphabetical WebSocket list


Jeffiy:
cc @quozd @devkhan @vladen

PR #231, opened 2015-12-02
Add MaterialSkin

wk-j:
https://github.com/IgnaceMaes/MaterialSkin


Jeffiy:
It seems the code didn't have a good format and test, cc  @quozd @devkhan @vladen

quozd:
We don't have any Material Design project for WinForms yet, so I think we can add it though it doesn't comply with our standards.

PR #235, opened 2015-12-06
Added Flurl to HTTP list

tmenier:
Thanks for your consideration!


pleb:
@tmenier thanks contributing. Looks like you may have a line ending issue, because +665 −664 changes. Can you please fix and we'll be happy to consider the PR

PR #238, opened 2015-12-07
added Flurl to HTTP

tmenier:


Jeffiy:
@tmenier You should make a PR for a commit, can you fix it ? thanks.

PR #243, opened 2015-12-18
Add link to Protogame

hach-que:
Protogame is a cross-platform game engine written on top of MonoGame.  It provides many higher level features such as entity management that are not included in MonoGame.

Website: https://protogame.org/


pleb:
My review:

It's still beta so the only rule potentially being broken would be "Stable". 

Here's the link to the repo https://github.com/hach-que/Protogame

I give a 1+

hach-que:
I'm still heavily reworking the project creation process, so if you want to hold off merging the PR until that stabilizes then I'm fine with that.

pleb:
@hach-que I'd like to see what the other reviewers think before merging. My position is, given this is a niche area, a product like yours is most likely still of use to the community. However, I do have one question. On the repo page, I see there's no issues tab. How would a user alert you to a problem?

hach-que:
My intent is to drive issues through this form https://code.redpointgames.com.au/maniphest/task/edit/form/1/ and in future to manage code contributions through there (unfortunately you can't turn off the Pull Requests tab on GitHub).  Putting this into the documentation / website is again on the to do list.

quozd:
Hi @hach-que, this project is interesting, however, we can't merge it right now. We can skip stability rules for your case, but users should have clean way to report issues. While you don't use github's issues page, you should add link to fabricator to the repo and website, so that users won't be confused. After that we'll merge this PR. Thanks

PR #245, opened 2015-12-23
Added Bundle Transformer to Assets

Taritsyn:
[Bundle Transformer](http://bundletransformer.codeplex.com/) - a modular extension for System.Web.Optimization (also known as the [Microsoft ASP.NET Web Optimization Framework](http://aspnetoptimization.codeplex.com/)). Its modules supports LESS, Sass, CoffeeScript, TypeScript, Mustache, Handlebars, Autoprefixer along with a bunch of different JS and CSS minifiers.


Jeffiy:
@Taritsyn Thanks.

PR #247, opened 2015-12-26
Added Minification category (second attempt)

Taritsyn:
I propose to create the “Minification” category and add the following projects:

* [Microsoft Ajax Minifier](http://ajaxmin.codeplex.com/) - Contains JS and CSS minifiers which have a highest performance, because its have been specifically designed for .NET. Optionally produce Source Maps for JS code.
* [Web Markup Minifier](http://webmarkupmin.codeplex.com/) - .NET library that contains a set of markup minifiers. The objective of this project is to improve the performance of web applications by reducing the size of HTML, XHTML and XML code.


Jeffiy:
@Taritsyn Thanks.

PR #249, opened 2016-01-08
Adding DotImaging

dajuric:
DotImaging - .NET array as imaging object
The framework sets focus on .NET native array as primary imaging object, offers extensibility support via extensions, and provides unified platform-abstract imaging IO API.


Jeffiy:
LGTM, but it seems does not have enough test.

dajuric:
OK, if you are referring to the Test folder it does not include them. The library is mature enough judging by the number of package downloads: https://www.nuget.org/packages?q=dotimaging

Should I add unit tests and then notify you or (this may take a while)... ?

Jeffiy:
@dajuric Fine, it could be better if you add the unit tests, thanks.

dajuric:
Thank you.

PR #251, opened 2016-01-13
Adding Infer.NET to Machine Learning

felipegtx:
Infer.NET provides the state-of-the-art message-passing algorithms, graph algorithms, and linear algebra routines needed by modern machine learning applications


Jeffiy:
@felipegtx Thanks, but it seems not active now, the last update was November 25 2014, and I can't find a way to report issue too.

 /cc @quozd @pleb

pleb:
@Jeffiy This might be an exception to the rule. It's an MS research project, so I believe these are driven by funding/grants then published. I guess as it's more of an academia project than a traditional project, the issue reporting could be overlooked. I wonder if it would be ok to accept PRs like this if they  are given some form of "flair" so people can easily distinguish between them.

felipegtx:
@pleb I couldn't have said it better. 

@Jeffiy from my perspective as a user of Infer.NET, IMHO there's nothing like it available for the platform today, thus providing it with the awesomeness points it need to make into the list. 

Maybe a `research` *flair* as suggested by @pleb would suffice.

quozd:
lets go with `research` tag. someone should update guidelines for this btw :)

felipegtx:
@quozd Just did it. Please take a look and let me know if you want me to change anything else. :+1:

pleb:
Awesome work everyone!

quozd:
@felipegtx :+1:

PR #262, opened 2016-02-23
Added DateTimeExtensions

kappy:
DateTime/DateTimeExtensions - [https://github.com/kappy/DateTimeExtensions](https://github.com/kappy/DateTimeExtensions)

Common Date Time operations on the form of extensions to System.DateTime, including holidays and working days calculations on several culture locales.


Jeffiy:
@kappy Thanks.

PR #264, opened 2016-02-24
Added FlatMapper

kappy:
Misc/FlatMapper - [https://github.com/kappy/FlatMapper](https://github.com/kappy/FlatMapper)

FlatMapper is a library to import and export data from and to plain text files in a Linq compatible way.


Jeffiy:
@kappy Thanks.

PR #268, opened 2016-03-20
Added dotNetify

dsuryd:
Web Frameworks/dotNetify - https://github.com/dsuryd/dotNetify

DotNetify is a simple, lightweight, yet powerful way to build modern .NET web applications


Jeffiy:
@dsuryd , thanks, It's looks great, but I'm not sure about it whether`Stable`, because it seems just to be created at 4 months ago.

/cc @quozd

devkhan:
It looks pretty nice, but I agree with @Jeffiy that it is pretty new, we should give it some time.

quozd:
Closed. Please, reopen it or create a new PR when the project will be more mature. Thanks

PR #273, opened 2016-04-06
Add Testing/ExpectedObjects

wellwind:
[Expected Objects](https://github.com/derekgreer/expectedObjects)

A testing library for creating Expected Objects. The Expected Object pattern is a technique involving the encapsulation of test-specific logic within a specialized type designed to compare its configured state against that of another object.


Jeffiy:
@wellwind I'm not sure about whether is it stable, and it's last updated at Jul 24, 2015.

/cc @devkhan @quozd

devkhan:
@Jeffiy Also, it doesn't look under active development either. I think we should let it go.

wellwind:
I'm still using it but maybe it's now awesome enough...let's close it.

PR #274, opened 2016-04-06
Add ReactJS.NET

wellwind:
Visual Studio Plugins/[ReactJS.NET](https://github.com/reactjs/React.NET)

ReactJS.NET is a library that makes it easier to use Babel along with Facebook's React and JSX from C#.

(I'm not sure is this good for "Visual Studio Plugins", maybe create another category like "web development" or "front end" is better)


devkhan:
I think this should be in the miscellaneous category as it is not exactly a VS plugin. @quozd What do you think?

devkhan:
@wellwind Can you rebase this?

PR #286, opened 2016-05-07
Add ReactJS.NET

wellwind:
Visual Studio Plugins/[ReactJS.NET](https://github.com/reactjs/React.NET)

ReactJS.NET is a library that makes it easier to use Babel along with Facebook's React and JSX from C#.


Jeffiy:
@wellwind Thanks.

PR #294, opened 2016-06-06
Add Zebus

ablanchet:
Distributed Computing/Zebus - https://github.com/Abc-Arbitrage/Zebus

Zebus is a lightweight, highly versatile, peer to peer service bus, built with CQRS principles in mind. It allows applications to communicate with each other in a fast and easy manner. Most of the complexity is hidden in the library and you can focus on writing code that matters to you, not debugging messaging code.


Jeffiy:
@ablanchet Thanks.

PR #305, opened 2016-06-29
Adding Croniton

leosperry:
Scheduling/Chroniton- [https://github.com/leosperry/Chroniton](https://github.com/leosperry/Chroniton)

A simple, fully integrable, and customizable library for running strongly typed jobs (tasks) on schedules.


Jeffiy:
@leosperry Thanks, but I doubt whether it `Stable`, because it's very new, so we should wait it becoming stable.

leosperry:
I appreciate the honesty, and you are absolutely right; it is new. I am currently looking into a continuous integration solution for my open source project. I do have a fair amount of code coverage in my tests. My question to you is "What do you consider stable?". The answer to that question will give me a goal to shoot for.

Jeffiy:
It's hard to see, but it's good that you have many tests. I know that `new != unstable`, but because it's new, this project didn't have many people to use it, maybe it had some obscure bugs that should someone to find it by really using.

@devkhan do you have any suggest?

devkhan:
I think we should give it sometime to prosper.

leosperry:
I've just added CI with tests running and pushed out Nuget packages.

pleb:
@leosperry cool library buddy!

I wonder what the others think now that you have unit tests?

PR #306, opened 2016-06-30
Added Magick.NET

dlemstra:
Image Processing/Magick.NET - [https://magick.codeplex.com/](https://magick.codeplex.com/)

ImageMagick is a powerful image manipulation library that supports over 100 major file formats (not including sub-formats). With Magick.NET you can use ImageMagick in your .NET application without having to install ImageMagick on your server or desktop.


Jeffiy:
@dlemstra Thanks.

PR #308, opened 2016-06-30
Fixed link for Metrics.NET

dlemstra:
Code Analysis and Metrics/Metrics-Net

It looks like this project has moved to a new URL. The new maintainer posted a message about it here: https://github.com/etishor/Metrics.NET/issues/122#issuecomment-184707498. The previous owner of the repo where this was forked from has passed away: https://github.com/etishor/Metrics.NET/issues/122#issuecomment-175175100. The NuGet package (https://www.nuget.org/packages/Metrics.NET/0.3.7) now also uses this url.


Jeffiy:
@dlemstra It's a bad news. 

The url of this project is not correct, should be `https://github.com/Recognos/Metrics.NET`, please update it.

PR #309, opened 2016-06-30
Fixed SmartStore link

dlemstra:
E-Commerce and Payments/SmartStoreNET

Fixed the link. It should end with en/net instead, also using https instead of http.


Jeffiy:
@dlemstra Thanks.

PR #316, opened 2016-07-25
Fix 404

felipemfp:
@danielcrenna has moved all his projects into [danielcrenna/vault](https://github.com/danielcrenna/vault).


Jeffiy:
@felipemfp Thanks

PR #321, opened 2016-07-27
Add Stuntman

kendaleiv:
Authentication and Authorization/Stuntman - [http://rimdev.io/stuntman/](http://rimdev.io/stuntman/)

Stuntman is a library for impersonating users during development leveraging .NET Claims Identity.


pleb:
Cool library

PR #323, opened 2016-07-28
Adding FsShelter reference

et1975:
Distributed Computing/FsShelter - [FsShelter](https://github.com/Prolucid/FsShelter)
F# library for authoring Apache Storm topologies - high level abstractions for distributed event stream processing using statically typed streams.


pleb:
I struggled with satisfying "Generally useful to the community". However, considering this is a niche area I think the rule does not need to apply here

PR #328, opened 2016-08-10
Added Netling

AsifMushtaq:
Test/Netling - https://github.com/hallatore/Netling

Netling includes both a WPF and Console client and is cleanly factored with a Core project that does all the work. With the WPF version you do test and then optionally mark that test as a baseline. Then you can make small changes as you like and do a quick retest. You'll get red (bad) or green (good) results if things get better. This should probably be adjusted to ensure it is visible for those with red-green color blindness. Regardless, it's a nice clean UI and definitely something you'll want to through into your utilities folder and call upon often!

687474703a2f2f692e696d6775722e636f6d2f687054624873712e706e67

687474703a2f2f692e696d6775722e636f6d2f386762506b784b2e706e67


pleb:
Hi @AsifMushtaq. Thanks for the contribute. Keep up the good work!

PR #362, opened 2016-08-30
Added Scrutor

khellang:
IoC/Scrutor - [https://github.com/khellang/Scrutor](https://github.com/khellang/Scrutor)

Assembly scanning extensions for Microsoft.Extensions.DependencyInjection

Should probably be merged after #359

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-﻿# Awesome .NET!
+# Awesome .NET!
 
 [![Awesome](https://cdn.rawgit.com/sindresorhus/awesome/d7305f38d29fed78fa85652e3a63e154dd8e8829/media/badge.svg)](https://github.com/sindresorhus/awesome) 
 [![Build Status](https://api.travis-ci.org/quozd/awesome-dotnet.svg)](https://travis-ci.org/quozd/awesome-dotnet)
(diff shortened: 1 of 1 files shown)


devkhan:
Is this mature enough?

khellang:
Oh. I didn't know there was a limit on how new the project was. I think it qualifies based on the current standards:

> To stay on the list, projects should follow these quality standards:
> 
> - Generally useful to the community
> - Actively maintained (even if that just means acknowledging open issues when they arise)
> - Stable
> - Documented
> - Tests

Maybe you should add an extra point to CONTRIBUTING.md?

BTW, the project has more stars, watchers, forks, contributors, commits and/or releases than several projects on the list already:

 - https://github.com/ajorkowski/NodeAssets
 - https://github.com/IdentityModel/IdentityModel / https://github.com/IdentityModel/IdentityModelv2
 - https://github.com/danielcrenna/vault/tree/master/oauth
 - https://github.com/logibit/logibit.hawk/
 - https://github.com/spouliot/gendarme
 - https://github.com/exceptionless/Exceptionless.DateTimeExtensions
 - https://github.com/Abc-Arbitrage/Zebus

And I only went through less than half the list. I think there needs to be some consistency if you're going to enforce something like this. How old is "old enough"?

IMO, it's really sad that newer projects can't get on the list. They're probably the ones that need the "publicity" the most... 😢

devkhan:
I apologize if that sounded pretentious or elitist, but I didn't looked at the project and was just asking you about the stability and reliability of the project as you're the owner of the repo. 

I agree that new, nice projects need to be mentioned, my statement was just to draw attention of other collaborators.

pleb:
Hi @khellang. 

Maturity can see as a combination of all categories. The primary reason for this is that a project lacking in any category is most likely not mature. However, exceptions are made on a case-by-case basis. 

I've merged this project because it's a feature addition to .net core; a project which is new and on the path to maturity itself!

You're right about new projects needing publicity, however, it's a tricky line to draw and follow. After all, this list is for finding `awesome` dotnet projects, but how does one define a new project as awesome? New projects may seem or sound awesome but are they?

Maybe the answer here might be a second list of `new` and probably `awesome` projects? Maybe new projects can graduate from the new list once they have proven themselves?  Thoughts?

Wade

PR #380, opened 2016-09-09
add NDepend

psmacchia:


pleb:
A heading was added which isn't needed

PR #387, opened 2016-09-11
add Simple.Data

MNie:
ORM/Simple.Data - http://simplefx.org/simpledata/docs/index.html

Simple.Data is a lightweight framework that uses the dynamic features of .NET 4 to provide an expressive, ORM-ish way of accessing and manipulating data without any of the code pre-generation and boilerplate required by other frameworks.

Add simple.data to orm frameworks


pleb:
@MNie looks like the project is abandoned. Do you agree?

MNie:
@pleb "Latest commit 65ff01e  on Mar 31, 2014" Yup it looks like, so I will close a pull request.

PR #402, opened 2016-09-27
Add Managed Windows API

hgupta9:
UI Automation/Managed Windows API - [sourceforge](http://mwinapi.sourceforge.net/)

Introspect and automate third-party Windows / VC++ applications by controlling the UI elements of the app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ Thanks to all [contributors](https://github.com/quozd/awesome-dotnet/graphs/cont
   * [Testing](#testing)
   * [Tools](#tools)
   * [Trading](#trading)
+  * [UI Automation](#ui-automation)
   * [Visual Studio Plugins](#visual-studio-plugins)
   * [Web Frameworks](#web-frameworks)
   * [Web Servers](#web-servers)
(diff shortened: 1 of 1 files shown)


pleb:
@hgupta9 I'll leave this for the community. I couldn't find any unit tests, but because it's a niche product I think it could be overlooked

hgupta9:
Its very reliable and I have made some additions too. Perhaps I'll upload it sometime.

pleb:
@hgupta9 

If you have a version with additions, my guess is the SourceForge version is abandon ware?

Did you make to make GH repo for it and repath the link? The reason I ask is so that people submit issues and PRs etc. Thoughts?

SlowLogicBoy:
I would love to have this linked somewhere, maybe it's not really suited here, but is there another .net list that would allow this to be added? If so we can add it there otherwise maybe we could have some section like untrusted/obsolete (but this would defeat the purpose of this list I guess). 
Anyway thanks @hgupta9 for posting this, I always Pinvoked Native/Unamanaged Windows API, I guess I will no more.

zihotki:
It's a very niche but when you need to work with Windows API it's really awesome

hgupta9:
@pleb Should I create a GH repo for this since its abandoned by the original author? I could upload my additions too?

PR #405, opened 2016-09-28
Adding .NET Core in Action to Books section

dmetzgar:
Please consider including my book, .NET Core in Action, into your Books section. The first chapter is free. Although I'm not sure if waiting 1 year applies to books (or .NET Core).


dmetzgar:
Thanks!

PR #412, opened 2016-10-06
Add Rider to IDE

MisterX2000:
Rider - https://www.jetbrains.com/rider

A cross-platform C# IDE based on the IntelliJ platform and ReSharper.
Currently in private Early Access.


devkhan:
It's a nice addition, but since it is still in EAP, I don't know if it qualifies as "stable".
Let's see what the other maintainers think.

pleb:
I agree with @devkhan. But, what an exciting project right? I wonder, given the official tooling for netcore is still beta, if we could allow an exception for this?

devkhan:
I'm completely in favor. Go ahead.

On Oct 17, 2016 4:51 AM, Wade Baglin <notifications@github.com> wrote:

I agree with @devkhan<https://github.com/devkhan>. But, what an exciting project right? I wonder, given the official tooling for netcore is still beta, if we could allow an exception for this?

PR #433, opened 2016-10-20
Add PubNub to Distributed Computing heading.

adamskt:
Distributed Computing/PubNub - [PubNub](https://www.pubnub.com/)

Realtime apps made simple. APIs for developers building secure realtime mobile, web, and IoT applications. [C# SDKs here](https://www.pubnub.com/docs/c-sharp-net-c-sharp/pubnub-c-sharp-sdk). **[Free]**


zihotki:
It'd be better to change the description because currently it's completely not possible to identify what is it about. Also, since it's a service bus, it's better to put it into Queue section where most of other service buses reside. Also, the library is free but it requires PubNub account to be used, which also has a free plan, but it's not free, it's freemium

zihotki:
Closing the pull request because of lack of interest from the contributor and raised concerns

PR #437, opened 2016-10-27
P Language and modification in Compiler header

devkhan:
Compilers, Transpilers and Languages/P - [P](https://github.com/p-org/P)

> P is a language for asynchronous event-driven programming. P allows the programmer to specify the system as a collection of interacting state machines, which communicate with each other using events.

This seems like a nice project, but I wasn't sure about where to put it, so I'm proposing the change in `Compiler` category, i.e. rename it to `Compilers, Transpilers and Languages`.

Moderators, please give your input.


pleb:
I like the name up. However, is P a dotnet related product?

PR #441, opened 2016-11-06
Added Jot - a library for managing application state (a better altern…

anakic:
Misc/Jot - [https://github.com/anakic/Jot](https://github.com/anakic/Jot)

PROJECT_DESCRIPTION

Jot is a .NET library for managing application state. It is designed to save and restore all application data that isn't domain related. 

For example, in an accounting application, it would be used for storing and reapplying all data that's not related to accounting, so it would be used for: window locations and sizes, last entered data (e.g. username), configuration settings and user preferences.

It substantially reduces the amount of code needed to make object properties "survive" an application shutdown. If integrated with an IOC container, making a property "persistent" is as easy as decorating it with a [Trackable] attribute.

### DIFF
--- a/README.md
+++ b/README.md
@@ -590,6 +590,7 @@ metadata in media files, including video, audio, and photo formats
 * [MediatR](https://github.com/jbogard/MediatR) - Simple, unambitious mediator implementation in .NET 
 * [Warden](https://github.com/warden-stack/Warden) - Define "health checks" for your applications, resources and infrastructure. Keep your Warden on the watch 
 * [ByteSize](https://github.com/omar/ByteSize) - ByteSize is a utility class that makes byte size representation in code easier by removing ambiguity of the value being represented. ByteSize is to bytes what System.TimeSpan is to time.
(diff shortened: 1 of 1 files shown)


anakic:
Crickets... Can anyone tell me if the readme file is confusing? Should I trim it down, and put the details in the wiki? Is it unclear what the library does, or if it's just not interesting (personally, I think it's spiffy:))?

zihotki:
The project is small but that's limited by the project's scope. Also, there are no alternatives for it like I checked. Also it can be used to easily save user settings. So considering the scope and functionality, given that tests are there, and that it may not look awesome at the first look but it's definitely awesome'ish, I'm merging it.

anakic:
Yaay:) True, it's not a huge library, but it's useful for practically any desktop application, and the alternative (using .settings files) isn't pretty. With the appropriate storage implementations (e.g. session, cookie or asp.net user profile) I can see it being fairly useful in web applications too. In any case, really glad it's been deemed awesome'ish, made my day!!

PR #447, opened 2016-11-24
Add MailBody

doxakis:
Mail/MailBody - [https://github.com/doxakis/MailBody](https://github.com/doxakis/MailBody)

Create transactional email with a fluent interface (.net)

### DIFF
--- a/README.md
+++ b/README.md
@@ -543,6 +543,7 @@ Thanks to all [contributors](https://github.com/quozd/awesome-dotnet/graphs/cont
 * [MailKit](https://github.com/jstedfast/MailKit) - A complete cross-platform mail stack including IMAP, POP3, SMTP, authentication and more. Built on top of MimeKit.
 * [MimeKit](https://github.com/jstedfast/MimeKit) - A cross-platform .NET MIME creation and parser library with support for S/MIME, PGP, TNEF and Unix mbox spools.
 * [PreMailer.Net](https://github.com/milkshakesoftware/PreMailer.Net) - C# library that moves your stylesheets to inline style attributes, for maximum compatibility with E-mail clients.
+* [MailBody](https://github.com/doxakis/MailBody) - Create transactional email with a fluent interface (.net)
 
 ## Mathematics
 


zihotki:
But what's awesome in it?

doxakis:
Hi,

It makes it easier to create (and maintain) transactional email.
Instead of dealing with html code, you can use the fluent interface written in c#.

Here is some arguments:

- Based on a template tested on all major email clients (https://github.com/leemunroe/responsive-html-email-template)
- Great for rapid development
- Quick examples to copy/paste and adapt for your needs (Ex: Email Address Confirmation, Password reset, Order confirmation, Notification, etc.)
- Extensible (Allow custom template by overriding the default one)
- You can also add custom html. See this example: https://github.com/doxakis/MailBody#custom-theme--raw-html
- Short and sweet syntax:

```
var body = MailBody
    .CreateBody()
    .Paragraph("Please confirm your email address by clicking the link below.")
    .Paragraph("We may need to send you critical information about our service and it is important that we have an accurate email address.")
    .Button("https://example.com/", "Confirm Email Address")
    .Paragraph("— [Insert company name here]")
    .ToString();
```

- Another way to create your email:

```
var body = MailBody.CreateBody();

body.Paragraph("Hi,")
    .Paragraph("First paragraph..");

// Your code

body.Button("https://www.example.com/", "First button");
body.Paragraph("Another paragraph..");

// Your code

body.Button("https://www.example.com/", "Second button")
    .Paragraph("— [Insert company name here]");

var htmlBody = body.ToString();
```

Thanks!

zihotki:
Let me be a bad policeman, I still don't see any awesomenes comparing to Razor Engine, it provides everything this utility has and even more than that. And also I don't see anything what is transactional there.

doxakis:
Hi,
You can read more about transaction email here : https://sendgrid.com/use-cases/transactional-email/

It is not replacing the Razor Engine. It only focus on one use case : make it easier to write transaction email.

You can write email with far less line of code.
It is easier to maintain and it focus on readability.

In your razor view, you would have to write lot of code to support many email client.
For example, when you add a button, it actually add a "table" with "a" element. (http://www.industrydive.com/news/post/how-to-make-html-email-buttons-that-rock/)
So it works on most email client : gmail, yahoo, hotmail, outlook, etc.

zihotki:
Sorry, but Razor Engine completely covers that by - Master Templates and Partials. Button and other stuff can be easily moved into partials, main template can be put into master template, if you need to add a button to specific email - add it using the partial for buttons. And you get full control about the content, get syntax highlight, intellisense, variables escaping, way better flexibility, you can even make some adjustments to the emails at runtime if you add some sort of CMS for Razor Engine email templates. The library is good for specific scenarios but it's not awesome IMO.

pleb:
@doxakis I agree with all @zihotki points. However, I can see a library, such as this, would allow a dev to add simple email support in a programmatic way without too much up-front effort. I could see myself using this in a prototype stage or small project. Therefore, I acknowledge your product has a market but is it deserving to go on this list? 

Given how new the library is, and that many others have been rejected because of this, I cannot accept this PR. I do hope in a short time you will be able to resubmit this with a better outcome! :)

PR #446, opened 2016-11-21
adding App Metrics - https://github.com/alhardy/AppMetrics

alhardy:
Code Analysis and Metrics/App.Metrics - [App.Metrics]https://github.com/alhardy/AppMetrics)

App Metrics is an open-source and cross-platform .NET library used to record and report metrics within an application and reports it's health. https://alhardy.github.io/app-metrics-docs/

PR #448, opened 2016-11-27
adding App Metrics - https://github.com/alhardy/AppMetrics

alhardy:
CATEGORY/PROJECT_NAME - [PROJECT_LINK](PROJECT_LINK)

PROJECT_DESCRIPTION

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,7 @@ Thanks to all [contributors](https://github.com/quozd/awesome-dotnet/graphs/cont
 
 ## Code Analysis and Metrics
 
+* [App.Metrics](https://github.com/alhardy/AppMetrics) - App Metrics is an open-source and cross-platform .NET library used to record and report metrics within an application and reports it's health. See the [docs](https://alhardy.github.io/app-metrics-docs/) for me details.
 * [CodeMaid](http://www.codemaid.net/) - Visual studio extension to cleanup, dig through and simplify C#, C++, F#, VB, PHP, JSON, XAML, XML, ASP, HTML, CSS, LESS, SCSS, JavaScript and TypeScript coding.
 * [StyleCop](https://stylecop.codeplex.com/) - StyleCop analyzes C# source code to enforce a set of style and consistency rules
 * [Gendarme](https://github.com/spouliot/gendarme) - Extensible rule-based tool to find problems in .NET applications and libraries


alhardy:
@quozd re-created my PR https://github.com/quozd/awesome-dotnet/pull/446 to kick off another build due to failures. Looks like my branch is already up to date with master but getting a different build error now?

pleb:
@zihotki Thoughts?

zihotki:
I see that the library is a port of Metrics.Net to netstandard platform. It was refactored, cleaned and the project structure has been improved by a lot. The fork itself is new, but core code for metrics calculation is mostly the same. Alhardy made really good progress and improved project structure, all decisions seem to be very reasonable. 
Metrics.Net has been abandoned for a couple of years and only recently got some progress (I see that the ownership has been transferred to new person, more discussions, etc.), also that they don't have any visible roadmap and vision so it would take some time to settle everything there. 
If I was asked to choose between App Metrics and Metrics.Net I would go with 1st. As for including in the list, I would definitely include both of them.

quozd:
If Metrics.Net is abandoned, we shouldn't add it to the list. AppMetrics looks ok.

alhardy:
Metrics.NET isn't abandoned however there is not much progress being made there. I contributed [Owin.Metrics](https://www.nuget.org/packages/Owin.Metrics/) a while back, the project was very active back then with the original owner, I see the project very useful so am trying to keep it alive by freshening it up and adding new features.

Before porting I did attempt to get a .NET Standard build underway see https://github.com/etishor/Metrics.NET/issues/133 and https://github.com/Recognos/Metrics.NET/issues/32. I also created an ASP.NET Core full framework [solution](https://github.com/alhardy/aspnet-metrics) using Metrics.NET. However I really wanted:

- a cross platform solution but also keeping support for older versions of the .net framework
- keep the project going and adding new features, I have plans for plenty once I'm happy with the inital clean up
- clean up the solution to make it more maintainable, quicker to add features and follow a more modern api into the library

I've given credit to the original owner in the documentation and code files where required, this is mainly the sampling code that has been written. I've benchmarked this with other solutions, he's done an amazing job no need to touch that, the only thing there I'll be changing the the HdrHistogram implementation which [HdrHistgram.NET](https://github.com/HdrHistogram/HdrHistogram.NET/issues/17) which I'm current in the process of finishing a PR to port to .NET Standard

pleb:
Thanks @zihotki for that excellent summary. I think based on @zihotki's input and @alhardy's response, we all agree for the inclusion of this library. Given that Metrics.Net is not abandoned but has limited recent progress, we'll skip adding it at this time - which addresses @quozd well-founded concerns. If somebody else submits a PR for it in the future, we can tackle this outstanding question at that point.

PR #451, opened 2016-12-16
Add CodeJam

NN---:
Misc/CodeJam - [https://github.com/rsdn/CodeJam](https://github.com/rsdn/CodeJam)

PROJECT_DESCRIPTION

Set of handy reusable .NET components that can simplify your daily work and save your time when you copy and paste your favorite helper methods and classes from one project to another

### DIFF
--- a/README.md
+++ b/README.md
@@ -593,7 +593,7 @@ metadata in media files, including video, audio, and photo formats
 * [MediatR](https://github.com/jbogard/MediatR) - Simple, unambitious mediator implementation in .NET 
 * [Warden](https://github.com/warden-stack/Warden) - Define "health checks" for your applications, resources and infrastructure. Keep your Warden on the watch 
 * [ByteSize](https://github.com/omar/ByteSize) - ByteSize is a utility class that makes byte size representation in code easier by removing ambiguity of the value being represented. ByteSize is to bytes what System.TimeSpan is to time.
-* [CodeJam](https://github.com/rsdn/CodeJam) - Set of handy reusable .NET components, that can simplify your daily work and save your time when you copy and paste your favorite helper methods and classes from one project to another.
+* [CodeJam](https://github.com/rsdn/CodeJam) - Set of handy reusable .NET components that can simplify your daily work and save your time when you copy and paste your favorite helper methods and classes from one project to another.
 
 ## MVVM
 
--- a/README.md
+++ b/README.md
@@ -593,7 +593,7 @@ metadata in media files, including video, audio, and photo formats
 * [MediatR](https://github.com/jbogard/MediatR) - Simple, unambitious mediator implementation in .NET 
 * [Warden](https://github.com/warden-stack/Warden) - Define "health checks" for your applications, resources and infrastructure. Keep your Warden on the watch 
 * [ByteSize](https://github.com/omar/ByteSize) - ByteSize is a utility class that makes byte size representation in code easier by removing ambiguity of the value being represented. ByteSize is to bytes what System.TimeSpan is to time.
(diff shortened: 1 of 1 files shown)


NN---:
Anyone can explain why the check fail ?

```
  1. [L321]  https://slimdx.org/ SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B: certificate verify failed 
```

zihotki:
@NN--- it fails because the certificate at https://slimdx.org/ has expired and unfortunately the site doesn't have an http version.

zihotki:
I have complex feeling regarding it - there are a lot of such all-in-one utility libraries and to know what's there you have to invest a lot of time. Also at some point you may find that you need another piece which is in different all-in-one library so you add it too for only one or two functions. And you can endup with big mess. On my practice and considering functionality provided by the library when you need code like that then it's most of the time a code-and-forget code. Such code is just easier to stackoverflow or find a gist and copy it into your project given that it can be easily done using [Developer Assistant] add-in (https://marketplace.visualstudio.com/items?itemName=OneCodeTeam.DeveloperAssistant-13032).

andrewvk:
The main difference between this library and code-and-forget code is quality - significant efforts invested in the performance, readability, flexibility, documentation and so on.
So, using this library is much better than copypasting dirty code from project to project over and over again.

zihotki:
@andrewvk Is it? Does it really matter if code executes in 1 or 2 seconds if code is not in hot path? I doubt that. If it's in hot path and you care about performance - it'd be optimized and performance tested. And most of code you can easily find is good enough. 
Documentation? What I see there is just xml documentation comments but that's can't be called a good documentation. In many cases they may even be called redundant because method names and signatures are self-describing. Sorry, but I can't call a documentation generated from xml comments and uploaded in github wiki a good documentation. Good documentation should be like [this](https://github.com/serilog/serilog/wiki). As for utility functions the best documentation is the source code of the function.
Also in case if you are writing high-performance application it's even better to copy the code because in that case it'll be owned by a team, which in it's turn will allow to do optimizations or refactorings in place.
My point is not that it's not performant or it's bad but that it's not awesome and small.

andrewvk:
1) Performance is not the only impotant thing. Code readability, good identifiers, Resharper annotations,  good test coverage is equally important. And why you should do optimize code if you can get it fo free? Why you use .NET FW, if you can write it from scratch or copy sources in every project?
2) Xml doc comments is not the only doc. See https://github.com/rsdn/CodeJam/tree/master/PerfTests/docs for example. And this is ongoing process.
3) Of course, you can get any piece of library code and copy to own project, and quality of this code, on average, is much better, than pieces from SO.

I think, your arguments becoming more and more subjective. OK, you personally like to copy and paste trivial pieces of code from SO or MS samples in every project, rather than use refined code from library. But this doesn't make library worst, isn't it?

zihotki:
Yes, I'm subjective, that's not even a question. Awesomeness is also very subjective topic. I spent some time looking at the code, what I saw and what I don't like is that huge part of the code is a very trivial code. It also has some parts which should belong to separate libararies like CSV parser, Mapper, DI (Services), Ranges, Assertions. There are a plenty of more robust, solid, feature-rich, and well-documented alternatives for them and they are already widely used. If you exclude those parts you endup with a very small amount of code and percentage of trivial code will be very high. No matter if it's tested, annotated, etc. it's still trivial code. And more I look at it, more convinced I become.

pleb:
Hmm. I'm unsure of this one too. On one hand, yes, this is a nice little helper library. On the other hand is it awesome? Probably not awesome enough. Maybe we throw over to the community to decide?

pleb:
Sorry @NN---, the community didn't really step up and support this.

PR #457, opened 2017-01-20
Add Jolt

emre1512:
HTTP/Jolt - [Jolt](https://github.com/emre1512/Jolt)

An easy to use http client for C#/.NET. Can do asynchronous GET, POST, PUT, DELETE, HEAD requests. Send/get JSON, files and web forms. Upload to and download from ftp and some other little features.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Awesome .NET!
+﻿# Awesome .NET!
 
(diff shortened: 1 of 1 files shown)


zihotki:
Too young and no tests, http client should be mature and have robust tests with high coverage to call it awesome. May be later

PR #465, opened 2017-02-11
Add ReadLine

tsolarin:
CLI/ReadLine - [https://github.com/tsolarin/readline](https://github.com/tsolarin/readline)

PROJECT_DESCRIPTION

ReadLine is a GNU Readline like library built in pure C#. It can serve as a drop in replacement for the inbuilt Console.ReadLine() and brings along with it some of the terminal goodness you get from unix shells, like command history navigation and tab auto completion.

It is cross platform and runs anywhere .NET is supported, targeting netstandard1.3 means that it can be used with .NET Core as well as the full .NET Framework.


zihotki:
The project is quite new and small but it has a value and it seems awesome to me, not sure what to do about it. Does anybody have an opinion?

jongalloway:
This project got a lot of positive attention on Hacker News lately (https://news.ycombinator.com/item?id=13615817). I recommend adding it.

PR #469, opened 2017-02-19
Add Expecto to testing section

AnthonyLloyd:
Testing/Expecto - [Expecto](https://github.com/haf/expecto)

Expecto is a smooth testing framework for F#, with APIs made for humans, giving strong testing methodologies to everyone. Unit testing, property based testing, performance testing and stress testing.

With Expecto you write tests as values. Tests can be composed, reduced, filtered, repeated and passed as values, because they are values. This gives the programmer a lot of leverage when writing tests.

Expecto comes with batteries included, but it's still open for extension due to its compositional model.


AnthonyLloyd:
Cool Thanks

PR #472, opened 2017-02-23
add Bunder and Minifier

temilaj:
Minification/Bunder and Minifier - [https://github.com/madskristensen/BundlerMinifier](https://github.com/madskristensen/BundlerMinifier)

Visual Studio extension that let's you configure bundling and minification of JS, CSS and HTML files.
instantly producing updated bundled or minified file output


zihotki:
Please use one PR - one project

PR #473, opened 2017-02-23
add Twilio dotnet library and new category

temilaj:
SMS/Twilio dotnet library - [twilio-dotnet](https://github.com/twilio/twilio-dotnet)

PROJECT_DESCRIPTION

A .NET Library for sending SMS and emails with Twilio. Supports .NET, ASP.NET, ASP.NET MVC and ASP.NET Web Pages (WebMatrix).


zihotki:
It seems like the repository is abandoned and active development is happening here - https://github.com/twilio/twilio-csharp . 
Also, we should probably create a wider category for SMS/Phone calls

temilaj:
Oh yeah, I see that. I'll include it in my next commit, and add the new category too.

PR #474, opened 2017-02-24
Added Objective Sharpie

lukasz-pyrzyk:
Tools/Objective Sharpie- [https://developer.xamarin.com/guides/cross-platform/macios/binding/objective-sharpie/](https://developer.xamarin.com/guides/cross-platform/macios/binding/objective-sharpie/)

PROJECT_DESCRIPTION
Objective Sharpie is a command line tool to help bootstrap the first pass of a binding. It works by parsing the header files of a native library to map the public API into the binding definition (a process that previously was manually done).

Objective Sharpie uses Clang parse header files, so the binding is as exact and thorough as possible. This can greatly reduce the time and effort it takes to produce a quality binding.


pleb:
Sorry @lukasz-pyrzyk, the community didn't really step up and support this.

PR #475, opened 2017-02-24
add Twilio C#/.NET library and new category

temilaj:
SMS and Phone Calls / Twilio csharp/dotnet library - [twilio-csharp](https://github.com/twilio/twilio-csharp)

PROJECT_DESCRIPTION

A C#/.NET Library for sending and receiving phone calls and text messages with Twilio. Supports .NET 3.5+ and .NET Standard 1.4.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ Thanks to all [contributors](https://github.com/quozd/awesome-dotnet/graphs/cont
   * [SDK and API Clients](#sdk-and-api-clients)
   * [Search](#search)
   * [Serialization](#serialization)
+  * [SMS and Phone Calls](#sms-and-phone-calls)
   * [State machines](#state-machines)
   * [Static Site Generators](#static-site-generators)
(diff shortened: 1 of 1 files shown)


yevster:
Are client libraries for commercial APIs within the guidelines of this project? This doesn't even look like commercial software, just libraries to use a commercial service.

zihotki:
@yevster On the one hand your concerns are correct, but on the other hand there is no way to create a completely free of charge service like this where you can make phone calls or send SMS from your site. There are some alternatives for Twilio but they don't have a good .Net clients, AFAIK. And the twilio .Net library looks very solid - very good fluent DSL, good test coverage, active community. 
But let's ask for more opinions

pleb:
I'm in favour of this

temilaj:
using the  API is free, and you do get one phone number. However, you have to pay for the voice and messaging services. I guess it would be difficult to make those free.

PR #476, opened 2017-02-25
Add Mediator.Net

mayuanyang:
CATEGORY/PROJECT_NAME - [Mediator.Net](https://github.com/mayuanyang/Mediator.Net)

PROJECT_DESCRIPTION

A simple mediator for .Net for sending command, publishing event and request response with pipelines supported


pleb:
Thoughts: Yes it's new. However, there aren't too many in this space. Hmm

mayuanyang:
is it anything that i can do or should i just wait?

pleb:
@mayuanyang wait a few days. If nobody else comments I'll merge it in for you

mayuanyang:
same as IoC, we got many in the list. MediatR is good and they got pipeline in recent 3.0, Mediator.Net was designed to have pipelines when it was born, in fact it focus on the mediator's middleware ecosystem, it has got many middleware side projects which can be pluged into the pipelines, e.g serilog, unitofwork, eventstore and messagequeue etc, you might notice that having the all is useful for domain driven design and event sourcing. Thought that both frameworks are mediator pattern, but the design focus is different

PR #487, opened 2017-03-24
Add C# in Depth to books list

jskeet:
Books/C# in Depth- [3rd edition](https://www.manning.com/books/c-sharp-in-depth-third-edition) and [4th edition](https://www.manning.com/books/c-sharp-in-depth-fourth-edition)

Deep dive into the details of the C# language.

If you'd prefer to either have only one link (which version?) or links as separate entries, let me know and I'll create a replacement PR.


pleb:
@jskeet yes, you broke the rules, but I don't care. I'm a big fan. Keep up the good work! Accepted

PR #491, opened 2017-03-31
Fix link to Compilers, Transpilers and Languages category

VPashkov:
There was a little typo in the anchor for Compilers, Transpilers and Languages category.
It was:
#compiler-transpilers-and-languages
But should be:
#compilers-transpilers-and-languages


zihotki:
Good catch!

PR #496, opened 2017-04-04
Added Giraffe project to README

dustinmoris:
Added Giraffe (functional ASP.NET Core web framework) to the README.


pleb:
@dustinmoris A link shouldn't appear under more than one heading. You'll need to pick the more prominent heading and update the PR. Thanks Wade.

dustinmoris:
Hi, thank you for the feedback. I simply followed NancyFx as a guiding example, which is also listed under API and Web Framework. In this case I think Web Framework makes more sense if I have to pick one. I'll update the PR.

PR #501, opened 2017-04-26
Constraint Logic Programming (CLP) and Parser Combinator (PC) libraries

MovGP0:
Adding libraries for Constraint Logic Programming (CLP)
* [Decider](https://github.com/lifebeyondfife/Decider)
* [Google OR-tools](https://developers.google.com/optimization/)
* [Microsoft Solver Foundation](https://msdn.microsoft.com/en-us/library/ff524509.aspx)
* [NSolver](http://www.cs.cityu.edu.hk/~hwchun/nsolver/)

Adding Parser Combinators
* [FParsec](http://www.quanttec.com/fparsec/) - F# specific Parser Combinator 
* [GosuParser](https://github.com/kthompson/GosuParser)

Renaming Section "Mathematics" to "Mathematics and Physics" and adding Quantity System, which handles physical quantities rather than just units. 

* [Quantity System Framework](https://quantitysystem.codeplex.com/) - Library for handling physical quantities and units

You can cherry-pick individual commits if required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ Thanks to all [contributors](https://github.com/quozd/awesome-dotnet/graphs/cont
   * [CMS](#cms)
   * [Code Analysis and Metrics](#code-analysis-and-metrics)
   * [Code Snippets](#code-snippets)
+  * [Constraint Logic Programming](#constraint-logic-programming)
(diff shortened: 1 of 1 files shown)


zihotki:
Could you please split it into separate pull requests? It's really hard to review and discuss when everything is in the same pile

PR #508, opened 2017-05-15
Remove TinyIoC since it seems no longer active?

SlowLogicBoy:
* [TinyIoC](https://github.com/grumpydev/TinyIoC) - Single-file, easy and cross-platform IoC container

Last commit was made 11 months ago, and issues are not acknowledged.

I still use this library, however I think it no longer meets quality standards to be in awesome-dotnet.


zihotki:
It's indeed abandoned and a lot of actuall issues so it's better to not to recommend it. As soon as it will become supported we'll gladly add it back.

PR #514, opened 2017-06-11
Add ResXManager (Visual Studio Extension)

tom-englert:
Visual Studio Plugins/ResX Resource Manager - [ResX Resource Manager](https://marketplace.visualstudio.com/items?itemName=TomEnglert.ResXManager)

The most popular free tool to localize all kind of applications with resx-based resources.


zihotki:
Both of https://github.com/quozd/awesome-dotnet/pull/515 and this pull request are actually about the same tool/library. According to the convention a tool/library can appear on the list only once. What would be a better place to put it? I think that going with VS extension will be better and we can add a note to the description saying that there is a standalone version available.

tom-englert:
Then I'll go with #515, "Visual Studio Plugins" doesn't say much about what a tools does.

PR #518, opened 2017-06-12
Add splash generator

tom-englert:
WFP/Splash Generator - [Splash Generator](https://github.com/tom-englert/SplashGenerator)

Automates the generation of the WPF splash screen bitmap from a WPF control.


pleb:
Sorry @tom-englert,

- Too small
- Possibly not awesome
- No of response from the community suggested the previous statements are false

Don't let this setback cause you to lose heart, and we look forward to your next contribute.

Thanks, Pleb

PR #521, opened 2017-06-20
Add WinFsp

billziss-gh:
File Systems/WinFsp - [https://github.com/billziss-gh/winfsp](https://github.com/billziss-gh/winfsp)

Windows File System Proxy - FUSE for Windows. A set of software components for Windows computers that allows the creation of user mode file systems that fully integrate with the OS. Includes .NET layer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ Thanks to all [contributors](https://github.com/quozd/awesome-dotnet/graphs/cont
   * [ETL](#etl)
   * [Exceptions](#exceptions)
   * [Extensions librairies](#extensions)
+  * [File Systems](#file-systems)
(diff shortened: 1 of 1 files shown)


pleb:
Sorry @billziss-gh,

- Possibly not awesome
- No of response from the community suggested the previous statement was false

Thanks, Pleb

billziss-gh:
@pleb, thanks, I will not lose heart.

WinFsp is an operating system component with a .NET binding; unlikely that it can match the "awesomeness" of .NET software. Thanks for the laugh :-D

Idea for future rejections: "Thank you for your contribution. We find that this project is not a good fit for our list."

PR #524, opened 2017-07-06
Add dotPeek decompiler

javithalion:
New item for decompiler list

Decompilation/dotPeek - [https://www.jetbrains.com/decompiler/](https://www.jetbrains.com/decompiler/)

DotPeek is a free-of-charge standalone tool based on ReSharper's bundled decompiler. It can reliably decompile any .NET assembly into equivalent C# or IL code. I can directly export decompiled code to Visual Studio projects. Navigate thru types and code is quite user friendly.


javithalion:
Added two commits by mistake. I just fixed that.

PR #529, opened 2017-07-14
Add XunitOrderer

fulls1z3:
Testing/XunitOrderer - [https://github.com/fulls1z3/xunit-orderer](https://github.com/fulls1z3/xunit-orderer)

Implementation of ITestCaseOrderer to enforcing xUnit to run the facts in strict order


pleb:
Sorry @fulls1z3 ,

- Possibly too new
- Possibly not awesome

Thanks, Pleb

fulls1z3:
@pleb thanks for the reply, but regarding the purpose of this util it's quite normal that it's too small (*ofc if we talk about just this project*).

If we talk about it's too new, that's right. Maybe it should reach to a larger community, get some feedback and restructured a bit according to this feedback before being sent to *awesome-dotnet*.

I'll consider these facts next time I'll make a contribution.

Thanks, and keep up the awesome work 👍

pleb:
@fulls1z3 I've added this https://github.com/quozd/awesome-dotnet/issues/545

fulls1z3:
@pleb thanks again for proposing the secondary list, that's really very kind (*and ofc awesome*) of you 😄 

Anyway, I'll consider your advice in this topic for any further contributions.

PR #533, opened 2017-07-20
Add Cashew

Jaxelr:
Cashew - https://github.com/joakimskoog/Cashew

A .NET library for caching responses easily with an HttpClient through an API that is simple and elegant yet powerful.


zihotki:
Although it's small and quite new, it fills the caching gap for HttpClient quite nicely, the quality of documentation and code is good so I'm going to include it

PR #550, opened 2017-08-31
Adding SqlKata

ahmad-moussawi:
Database/SqlKata - [SqlKata](http://sqlkata.vivida-apps.com) 

SqlKata - Elegant Sql Query Builder, that supports complex queries, joins, sub queries, nested where conditions, vendor engine targets and more


zihotki:
I think that ORM section will be a better fit for the framework. Databases section contains databases and core frameworks to work with databases but SqlKata is a query builder which works on top of standard Sql stack.

pleb:
I think this project is awesome because

1. Nice site
2. Nice documentation
3. Not that many query builders I know of written for dotnet

My suggestion would be a new category under DB - Database query builders. The reason for this suggestion is that, as @zihotki states, it's not really a great fit for Databases. But, I think the same can be said for ORMs. 

If we add this category, we should move another query builders listed.

Thoughts?

ahmad-moussawi:
Agree with @pleb, a new category for query builders would be more appropriate, since now it cannot be considered as ORM

pleb:
@ahmad-moussawi can you update the PR pls. I'll merge when placed under new cat

ahmad-moussawi:
@pleb, updated thanks.

PR #567, opened 2017-09-11
Add IoC \ TinyIoc

CrazyAlex25:
IoC \ TinyIoc https://github.com/grumpydev/TinyIoC
An easy to use, hassle free, Inversion of Control Container for small projects, libraries and beginners alike. Single file!

### DIFF
--- a/README.md
+++ b/README.md
@@ -539,6 +539,7 @@ To the extent possible under law, [Vitali Fokin](https://github.com/quozd) has w
 * [Microsoft.Extensions.DependencyInjection](https://github.com/aspnet/DependencyInjection) - The default IoC container for ASP.NET Core applications.
 * [Scrutor](https://github.com/khellang/Scrutor) - Assembly scanning extensions for Microsoft.Extensions.DependencyInjection.
 * [VS MEF](https://github.com/Microsoft/vs-mef) - Managed Extensibility Framework (MEF) implementation used by Visual Studio.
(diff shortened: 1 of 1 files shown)


SlowLogicBoy:
We just removed this because this is no longer supported.

CrazyAlex25:
So from this he did not stop working. I use this because there are no external relocations, just one file and easy use

SlowLogicBoy:
I also use this extensively

pleb:
Looks like it's not maintained anymore. I've used this a few times too.

@grumpydev say it isn't so?

kiquenet:
Any full real world samples using It?

CrazyAlex25:
TinyIOC can be used in PCL and Xamarin.

SlowLogicBoy:
Nancy uses TinyIoC as default IoC container

grumpydev:
Its not abandoned no, it's actively used as Nancy's default container for one. There are a few reasons it hasn't been updated for a while. Firstly, the last year or so, for various personal and professional reasons, I've had very little time for OSS work. Secondly, the version in Nancy I *think* has diverged from the version in the main repo, so I need to consolidate before I do much more work on it, and finally I have (thankfully I think) managed to avoid any of the .net core/standard/vnext stuff so far, and the main "issues" I see discussed are all around that, although Nancy itself supports netstandard with TinyIoC inside, so I think that issue is "ok", but referring back to the previous point.

After the consolidation I intend to work on the way the child containers work, as currently you have to make registrations directly in the child container, rather than being able to specify a "child lifetime" on the parent container (which would be more efficient and less code for the consumers).

pleb:
@grumpydev sounds awesome. And, thanks for the update.

PR #570, opened 2017-09-11
Misc \ HidLibrary

CrazyAlex25:
Misc \ HidLibrary https://github.com/mikeobrien/HidLibrary
This library enables you to enumerate and communicate with Hid compatible USB devices in .NET.


CrazyAlex25:
> NOTE: Support is VERY limited for this library. It is almost impossible to troubleshoot issues with so many devices and configurations. The community may be able to offer some assistance but you will largely be on your own.

The repository contains a note about support: the library does not implement all definitions of class descriptors from the standard.

CrazyAlex25:
This project saves a lot of time and effort when working with HID devices. I've tried similar projects, but this one has many useful things.

PR #571, opened 2017-09-11
Add new category

CrazyAlex25:
Add new category "Event aggregator and messenger". Move Mediator.Net and MediatR. Add new Rx Event Aggregator and TinyMessenger


CrazyAlex25:
I create new PR with new cat #571

PR #581, opened 2017-09-28
added utf8json library

zihotki:
Serialization/Utf8Json - [Utf8Json](https://github.com/neuecc/Utf8Json)

> Definitely Fastest and Zero Allocation JSON Serializer for C#(.NET, .NET Core, Unity and Xamarin), this serializer write/read directly to UTF8 binary so boostup performance.

This is a quite new and young (on github but I assume author was working for longer time behind the doors) serialization library but it has comprehensive tests and is quite well tested. [Benchmarks](https://gist.github.com/Tornhoof/00c5b5e73d46ea4710a5a4b3c714ba50) show that it at least as fast as Jil but has much less allocations. It's the most memory efficient library and it can even be compared to protobuf


neuecc:
I'm lib author, notice more details for benchmark.
I've ported that benchmakr and measure three patterns, `Object <-> byte[](Utf8)/Stream(Utf8)/String(UTF16)`.
https://github.com/neuecc/Jil

The important thing, what type do you want to input/output.
If target is UTF8(both byte[] and Stream), Utf8Json wins and memory allocated is extremely small.

I can not add myself to "awasome",
but if you are interested in a fast binary serializer, MessagePack for C# also...
https://github.com/neuecc/MessagePack-CSharp

PR #589, opened 2017-10-22
Add CliWrap

Tyrrrz:
CLI/CliWrap - [CliWrap](https://github.com/Tyrrrz/CliWrap)

CliWrap is a library that makes it easier to interact with command line interfaces. It provides a convenient wrapper around the target executable, allowing you to pass execution parameters and read the resulting output. The library can also handle errors reported by the underlying process, allows command cancellation and has both synchronous and asynchronous APIs.


Tyrrrz:
What is the criteria for project size?

zihotki:
I assume a project is small if the functionality it provides can be re-created in 1-3 days by average developer

Tyrrrz:
Fair enough. One can argue that 3 days not spent doing something an average developer can do, is 3 days spent doing something useful.

Jogai:
Thats super relative. Imagine making a poc that needs this little project, but the whole poc can be made in a week because of it. Than its a tremendous timesaver.

Tyrrrz:
Considering some recent changes, I doubt this is the kind of a project an average developer would do in 1-3 days. For example, there quite a few edge cases and race conditions that are covered [here](https://github.com/Tyrrrz/CliWrap/blob/949e257e2b7af8184600eb8b58209a7fb2094ab4/CliWrap/Cli.cs#L94) which might not be obvious to someone who hasn't had experience working with this before. And to add to what @Jogai said, the library is also already fully tested which means you won't need to.

PR #596, opened 2017-11-08
Add docopt in CLI section

kidchenko:
CLI/docopt- [docopt](https://github.com/docopt/docopt.net)

Isn't it awesome how CommandLineParser and PowerArgs generate help messages based on your code?!

Hell no! You know what's awesome? It's when the option parser is generated based on the beautiful help message that you write yourself! This way you don't need to write this stupid repeatable parser-code, and instead can write only the help message--the way you want it.

docopt.net helps you create most beautiful command-line interfaces easily:


kidchenko:
I am not sure why the build broke '-'

PR #605, opened 2017-11-25
Add posh-dotnet (PowerShell tab completion for the dotnet CLI)

bergmeister:
CLI/posh-dotnet - [posh-dotnet](https://github.com/bergmeister/posh-dotnet)

`PowerShell` tab completion for the [dotnet CLI](https://github.com/dotnet/cli).

This module enhances the tab completion of the dotnet cli 2.0 (which is disabled by default) and also provides support for the dotnet cli 1.0 (which had not tab completion at all)


zihotki:
It seems to me that `CLI` section is not a good place for it. In that section we have libraries to build CLI-based programs. `posh-dotnet` is a tool (or extension to PowerShell) to help us running commands, I'd rather put it in the `Tools` section

PR #622, opened 2018-01-27
add DeviceDetector.NET

totpero:
Misc/[DeviceDetector.NET](https://github.com/totpero/DeviceDetector.NET)

The Universal Device Detection library for .NET that parses User Agents and detects devices (desktop, tablet, mobile, tv, cars, console, etc.), clients (browsers, feed readers, media players, PIMs, ...), operating systems, brands and models.
This is a port of the popular PHP [device-detector](https://github.com/matomo-org/device-detector) library to C#. For the most part you can just follow the documentation for device-detector with no issue.


zihotki:
Given that it's a complete port with tests and original library is widely used I'm including it in the list

PR #637, opened 2018-03-21
Add NReco PivotData

VitaliyMF:
Business Intelligence / NReco PivotData - [NReco PivotData](https://www.nrecosite.com/pivot_data_library_net.aspx)

In-memory data aggregation/OLAP library, pivot tables generation (render to HTML, exports), ASP.NET pivot builder control **[$][Free for single-deployment/non-SaaS]**

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ To the extent possible under law, [Vitali Fokin](https://github.com/quozd) has w
   * [Authentication and Authorization](#authentication-and-authorization)
   * [Books](#books)
   * [Build Automation](#build-automation)
+  * [Business Intelligence and Reporting](#business-intelligence)
   * [Caching](#caching)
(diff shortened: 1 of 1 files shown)


pleb:
Sorry, @VitaliyMF,

- Community decision required
- No response from the community 😢 

Thanks, Pleb

VitaliyMF:
@pleb not sure that I understand what 'response from the community' means - currently [NReco.PivotData](https://www.nuget.org/packages/NReco.PivotData/) nuget package has 15k downloads and this is good tracking for rather niche component like that.

However, I'm ok with either way :-)

pleb:
Let's give it one more chance ;)

Jogai:
It does look awesome. Hard to judge if you dont really use it, but the project looks like its doing really well.

PR #644, opened 2018-03-28
Add Http/LittleRestClient

EldieTurner:
HTTP/LITTLERESTCLIENT- [LIttleRestClient](https://github.com/EldieTurner/LittleRestClient)

simple rest client compatible with .net standard 2.0


pleb:
Sorry, @EldieTurner,

- Possibly too new
- Possibly too small
- No response from the community suggested the previous statements were false

Thanks, Pleb

PR #652, opened 2018-04-22
add AppCfg.Net, a configuration framework for .NET developers

minhhungit:
Tools/AppCfg.Net - [AppCfg.Net](https://github.com/minhhungit/AppCfg.Net)

Add AppCfg.Net - A mini but powerful configuration framework for .NET developers

Support: 
- AppSetting datasource
- MSSQL Database datasource
It also support multi-tenancy.
Besides, AppCfg.Net also support immutable setting. This help us avoid changing setting value at runtime

### DIFF
--- a/README.md
+++ b/README.md
@@ -885,6 +885,7 @@ metadata in media files, including video, audio, and photo formats
 
 ## Tools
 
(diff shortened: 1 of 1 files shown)


pleb:
Sorry, @minhhungit,

- Too new
- Too small

Thanks, Pleb

PR #662, opened 2018-05-07
Add Cinchoo NACHA library

Cinchoo:
Misc/Cinchoo NACHA - [Cinchoo NACHA](https://github.com/Cinchoo/ChoETL.NACHA)

A NACHA library for .NET / c#


pleb:
Sorry, @Cinchoo,

- Possibly too small
- Possibly not awesome

Thanks, Pleb

PR #666, opened 2018-05-19
Add BEPUphysics

fxredeemer:
Game/BEPUphysics - [https://github.com/bepu/bepuphysics1](https://github.com/bepu/bepuphysics1)

BEPUphysics is a pure C# 3D physics library by BEPU. It's fast and has a bunch of cool features like constraints, terrain, static and instanced meshes, continuous collision detection, custom collision rules, vehicles, easy multithreading.


zihotki:
Hey, thanks for the PR. It looks like v1 of the engine is there only for historical records and all development effort is in v2 - https://github.com/bepu/bepuphysics2 . Could you please update the PR with the v2?

fxredeemer:
Hey @zihotki 
The reason I added the version 1 ist that the 2nd version is still in alpha. Should I anyways updtae the PR?

zihotki:
@fxredeemer I think it's better to do that way, eventually it'll hit the release. And we don't make decisions about projects based on whether it's in alpha stage or RTM. It's up to the user decide about whether it's stable for him

PR #680, opened 2018-06-17
Add GitMind

FantasticFiasco:
Git Tools/GitMind - [http://michael-reichenauer.github.io/GitMind/](http://michael-reichenauer.github.io/GitMind/)

GitMind is a visual Windows Git client, which intends to make it easier to use Git when using a branching model similar to e.g. GitFlow. The goal is to visualize the branch structure on screen in the same way as the you imagines the structure in your mind.


pleb:
Hi @FantasticFiasco,

According to the webpage

> Status
The current code status is a prototype which tries to give a preview of the base concepts

I feel this entry request fails the Quality stanard in the [contributing gude](https://github.com/quozd/awesome-dotnet/blob/master/CONTRIBUTING.md)

Cheers, Pleb

FantasticFiasco:
Ah, that is to bad. Although, I have been using it constantly for 6 months, can really recommend it,  and feels more like the beta label Google had years on end on GMail than actually being a "beta". Nevertheless, I respect your opinion to not include it in the list.

Cheers!

PR #711, opened 2018-10-05
addition of MicroMesh

mikebz:
Distributed Computing/MicroMesh - [https://github.com/ethoslending/micromesh/](https://github.com/ethoslending/micromesh/)

Lightweight Service Mesh written in .NET Core.

What is a Service Mesh?  From [https://www.nginx.com/blog/what-is-a-service-mesh/](https://www.nginx.com/blog/what-is-a-service-mesh/)
A service mesh is a configurable infrastructure layer for a microservices application. It makes communication between service instances flexible, reliable, and fast. The mesh provides service discovery, load balancing, encryption, authentication and authorization, support for the circuit breaker pattern, and other capabilities.


zihotki:
Hey, thank you very much for your PR. From what I was able to see from the code the project currently has simple implementations of requests routing, retry functionality, and logging. From my understanding of a service mesh pattern it should do much more than that (fault tolerance, rate limiting, load balancing, metrics, etc.). Otherwise it becomes very hard to differentiate it from a proxy. So for now I'd prefer to not to add it to the list and wait till it get's a bit more mature and feature rich.

mikebz:
Good questions, the project does have fault tolerance and load balancing - we are using it in production.  It does not have rate limiting and metrics, so I agree with that.  No problem with not accepting it - I was just looking to share it with other folks.  The current meshes that I've seen are not written in .NET and a lot of them are pretty complex.

PR #719, opened 2018-10-13
Add Guard

safakgur:
Misc/Guard - [https://github.com/safakgur/guard](https://github.com/safakgur/guard)

Guard is an intuitive argument validation library that takes advantage of almost all the features that were introduced in C# 7.2 to provide high performance, allocation-free fluent syntax possible. See the [design decisions](https://github.com/safakgur/guard/blob/dev/docs/design-decisions.md) for the motivation behind its API design.


safakgur:
Is there anything I can do on my end?

PR #725, opened 2018-10-27
Add Lemonad

inputfalken:
[Lemonad](https://github.com/inputfalken/Lemonad)

A functional **C#** library with data structures and functions whose main goal is to have an [declarative](https://en.wikipedia.org/wiki/Declarative_programming) approach to problem solving by using various [HOF's](https://en.wikipedia.org/wiki/Higher-order_function#C#) to achieve cleaner code bases & code which is easier to think about.


pleb:
Sorry, @inputfalken,

- Possibly too new
- No of response from the community suggested the previous statement is false

Thanks, Pleb

PR #735, opened 2018-11-29
Add a new websocket client library

Marfusios:
Webscoket/Websocket.Client - [PROJECT_LINK](https://github.com/Marfusios/websocket-client)

A wrapper over native C# class ClientWebSocket with built-in reconnection and error handling.


pleb:
Sorry, @Marfusios,

- Possibly too new
- No of response from the community suggested the previous statement is false

Thanks, Pleb

Issue #635, opened 2018-03-14
mutation testing frameworks/tools


skorzinetzki:
I like your list and I got here because I have been looking for a mutation testing framework / tools. unfortunately there is nothing listed right now. I think there is nothing to find right now. but maybe, at any time, we have something that should get listed up here.

zihotki:
What do you mean by that?

skorzinetzki:
What do I mean by mutation testing? Mutation testing can be described as a method, with which you mutate the production code (e.g. "if (a>b)" will become "if (a<b)") and then run your automated test suite. With each mutation at least one test should fail, so that this mutation is killed. Otherwise you may have a lack in your code coverage (or other reasons).
so this is a method to measure the quality of your test suite and goes beyond code coverage metrics.

One concrete tool might be Visual Mutator (https://visualmutator.github.io/web/). But this is not working with current Visual Studio versions. 
Working tools/frameworks for other languages are PIT (http://pitest.org/) for Java or Stryker (http://stryker-mutator.io/) for JavaScript.

zihotki:
Oh, I wasn't aware of such concept, my bad. I tried to see how much it's widespread and it looks to me that adoption of it is not widespread as well as there aren't many libraries/frameworks in .Net supporting it. So I'd rather avoid creating a new specific category and I'd suggest to put it to the Testing category where we have frameworks and libraries related to unit/integration/younameit testing.

skorzinetzki:
Yes, this is the best fit. I did not want to suggest a separate category, but, at any time when available, to list such a framework in the testing category. ;)
On Twitter I got into a conversation with one of the maintainers of Stryker, which I mentioned earlier in this issue. They are planning to start to develop a tool for c#. Sounds promising!

pleb:
@skorzinetzki interesting... PR with new category seems like the right next step

philippw:
The Stryker mutation testing framework has gained inital support for C#. See https://github.com/stryker-mutator/stryker-net.

skorzinetzki:
@philippw yes, I had a first try and it worked in my very simple example. I think it can be listed right now. Or do you wait until it has left it's preview status?

skorzinetzki:
Okay, I can create a pr today. :)

zihotki:
I'm closing the issue since all questions have been covered and PR has been merged. For now let's keep mutation testing under the Testing section, later on we can move it to separate if there will be a plenty of similar projects. Please reopen the issue if needed.

PR #736, opened 2018-11-30
added Stryker.NET as mutation testing framework to the testing category.

skorzinetzki:
Testing/Stryker.NET - [https://github.com/stryker-mutator/stryker-net](https://github.com/stryker-mutator/stryker-net)

Mutation testing for .NET Core projects

Solves issue #635. Because diverse tools and frameworks are listed in the testing category, I decided to add Stryker right here and not in a separate mutation testing category and therefore against the suggestion in the issue.


skorzinetzki:
I checked the travis build, seems like it is broken because of other links, not my newly added one.

PR #742, opened 2018-12-30
add HandyControl - JsonSettings - VSNotification

ghost1372:
GUI/[HandyControl](https://github.com/NaBian/HandyControl) - Contains some simple and commonly used WPF controls
GUI/ [Wpf.Notifications](https://github.com/Enterwell/Wpf.Notifications) - WPF notifications UI controls (as seen in VS Code)

Serialization/[JsonSettings](https://github.com/Nucs/JsonSettings) - Easiest way you'll ever write settings for your app. As readable Json or encrypted byte file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -486,6 +486,8 @@ To the extent possible under law, [Vitali Fokin](https://github.com/quozd) has w
 
 ## GUI
 
+* [HandyControl](https://github.com/NaBian/HandyControl) - Contains some simple and commonly used WPF controls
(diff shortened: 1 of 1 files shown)


zihotki:
Hey, thank you for the contribution. WPF controls look awesome and it would be a great to include them. But unfortunately I can't approve it right away because PR contains JsonSettings which it'll be better to review separately. The reason why the PR is required to have a single project, although it looks like it's more work for a contributor, is that it can be reviewed, discussed and merged independently.

PR #744, opened 2019-01-02
Add Proxmox VE Client Api

franklupo:
Add Proxmox VE Client Api


pleb:
Sorry, @franklupo,

- Missing tests
- Possibly too small
- Possibly not awesome

Thanks, Pleb

PR #745, opened 2019-01-08
Add Crypto-Notepad

Sigmanor:
Cryptography/Crypto-Notepad - [PROJECT_LINK](https://github.com/Crypto-Notepad/Crypto-Notepad)

Notepad-like simple text editor for Windows, where files are saved and loaded encrypted with AES algorithms.


zihotki:
Hey, thank you very much for the contribution.

I see that you added it to Cryptography section. That section currently contains cryptographic libraries but not tools using cryptography for some purposes. It looks to me that it'd be better to put it in Tools section since it doesn't fit IDE's as well and we don't have any section suitable for text editors. 

What do you think?

Sigmanor:
@zihotki Hey.
Yes, I think add in `Tools` category will be better.

pleb:
Agree with tools ;)

PR #754, opened 2019-01-29
Added new section: Strong Naming - starting with 'Strong Namer' library.

ElektroStudios:
Strong Naming/Strong Namer - [PROJECT_LINK](https://github.com/dsplaisted/strongnamer)

Automatically add strong names to referenced assemblies which do not already have a strong name. This will allow you to reference and use (NuGet packages with) assemblies which are not strong named from your projects that do use a strong name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,7 @@ To the extent possible under law, [Vitali Fokin](https://github.com/quozd) has w
   * [SMS and Phone Calls](#sms-and-phone-calls)
   * [State machines](#state-machines)
   * [Static Site Generators](#static-site-generators)
+  * [Strong Naming](#strong-naming)
   * [Style Guide](#style-guide)
   * [Template Engine](#template-engine)
   * [Testing](#testing)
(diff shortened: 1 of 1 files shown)


zihotki:
Hey, thank you for sharing the project, it's awesome. Although I'm not sure that we need a separate section for it given that we don't have any other candidates to put there as well. What do you think about listing it under `Tools` section since it's intention is to solve issue of strong naming, or `Misc`?

ElektroStudios:
Hi. Thanks for comment. You asked me what I think...

 - The premise for classifying something into sections, is to at least have one candidate per section. Then why it should be a problem to consider adding new section when there is a good (maybe the best) candidate to do that?.

 - Not initially having more than one candidate/entry can't be considered a reason to ignore the addition of important sections.

 - There are already sections with a single candidate/entry, like "Micro Framework" section.

 - Strong Naming is a very common practice for business scenarios and professional (commercial or not) libraries and applications, then, I think it deserve its own section. In fact, I consider that the "Misc" section is a design error, because every project listed there could be listed in more specific sections.

 - To do a little comparison: if exists a section for minifying code, being a feature that in overall it has no importance at all to be highlighted inside its own section (and only having two entries), then something that is related to file security like is Strong Naming such as versioning and naming protection, it must have a section... when there is at least one candidate for that.

Thanks for read.

pleb:
Thinking out loud. Its own section might encourage more libs in the space.

Issue #738, opened 2018-12-08
Can you add Easy Pdf


aleGuardiola:
I am working in this library and I think is awesome :). Can you take a look https://github.com/aleGuardiola/EasyPdf

Jogai:
The purpose of an awesome list on github is to supply a PR with your suggestion, see https://github.com/quozd/awesome-dotnet/blob/master/CONTRIBUTING.md

devSoheilAlizadeh:
PDF version has added with #758

Jogai:
That is something different entirely

PR #758, opened 2019-02-09
Add PDF version of Awesome .NET 

xsoheilalizadeh:
I've added PDF version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,9 @@ Contributions are always welcome! Please take a look at the [contribution guidel
 
 Thanks to all [contributors](https://github.com/quozd/awesome-dotnet/graphs/contributors), you're awesome and wouldn't be possible without you! The goal is to build a categorized community-driven collection of very well-known resources.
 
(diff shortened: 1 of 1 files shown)


zihotki:
Hey, Thank you for your PR, having a PDF version may be helpful. But what scares me is that it's manually generated and somebody need to maintain it. It'll be out of sync very quickly.

devSoheilAlizadeh:
@zihotki It's ok, I maintain it for till I prepare a process to make it automatically.

zihotki:
> @zihotki It's ok, I maintain it for till I prepare a process to make it automatically.

That's great but I still would like to discuss the need of the PDF first. It never occured to me that people will need PDF of the list and I wonder what's the purpose and what problem they're trying to solve.

By the way, the issue you linked #738 is about adding, I think, EasyPdf wrapper component https://github.com/aleGuardiola/EasyPdf . It wasn't a request to create a PDF version.

pleb:
Sorry @xsoheilalizadeh,

This entry isn't going to make the list.

Thanks, Pleb

PR #759, opened 2019-02-10
Workflow Core

danielgerlag:
Workflow Engines/Workflow Core - [WorkflowCore](https://github.com/danielgerlag/workflow-core)

Workflow Core is a light weight workflow engine targeting .NET Standard. Think: long running processes with multiple tasks that need to track state. It supports pluggable persistence and concurrency providers to allow for multi-node clusters.

** Also added `Workflow Engines` category

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,6 @@ To the extent possible under law, [Vitali Fokin](https://github.com/quozd) has w
   * [Web Servers](#web-servers)
   * [WebSocket](#websocket)
   * [Windows Services](#windows-services)
-  * [Workflow Engines](#workflow-engines)
   * [WPF](#wpf)
   * [Parser Library](#parser-library)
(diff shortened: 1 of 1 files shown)


zihotki:
Hey, Thank you very much for the PR. The project is definitely awesome but I prefer to add new categories after some discussion to keep categories under control. Let's see what other mods/people say

danielgerlag:
@zihotki which category would you suggest?

danielgerlag:
@zihotki ? @quozd ? somebody? anybody?

zihotki:
Hey, sorry for delay. I'd consider putting it into Scheduling category alongside the similar workflow frameworks. Or we can introduce a new category and move workflow frameworks from Scheduling to new category.

danielgerlag:
done 😄

PR #760, opened 2019-02-14
code formatter

hanabi1224:
dotnet code formatter - [codeformatter](https://github.com/dotnet/codeformatter)

Tool that uses Roslyn to automatically rewrite the source to follow netfx coding styles


zihotki:
Hey, thank you a lot for the contribution. It would have been a great addition but given that there are quite a lot of pending PRs, it has no official maintainers as it was mentioned in https://github.com/dotnet/corefx/issues/24185 , and the project's future is unclear as of now according to https://github.com/dotnet/codeformatter/issues/267 I don't think that adding it now will make any sense.

I'm personally eagerly looking for some automated formatter for C# and was but that one unfortunately can hardly be used. Considering my negative experience with it I'd prefer to wait till it will get some refactoring and facelift.

gep13:
What about this version of the tool, which seems to be the replacement for the above:

https://github.com/dotnet/format

It was recently released as a .Net Tool.

zihotki:
@gep13 just saw it from Scott Hanselman's post :) That one definitely need to be on the list!

PR #790, opened 2019-05-10
Some Changes

zandiarash:
There is some updates for GrandNode and  SmartStore


Jogai:
Seems to me it would be better to take smartstore off the awesome list. Even if they legally did nothing wrong its not so awesome...

Arashz22:
Yes, you are right dude but if SmartStore be in awesome list with that two link which I mentioned, it become more obvious that is legal fork.
and gradually programmers get away from that...

pleb:
Please remove SmartStoreNET

Arashz22:
> 
> 
> Please remove SmartStoreNET

Agree!

PR #805, opened 2019-06-21
Add Tangram framework to GUI section.

codemeow5:
GUI/Tangram - [https://github.com/TangramDev/Tangram](https://github.com/TangramDev/Tangram)

An amazing framework that embeds the entire chromium browser into the WinForm. Allow web pages and forms to be mixed.

We have a blog [post](https://blog.tangram.dev/a-new-way-to-define-desktop-software-using-web) to showcase our technology.


pleb:
Link doesn't work

PR #806, opened 2019-07-01
Add Simplify.DI

i4004:
IoC / Simplify.DI - [https://github.com/i4004/Simplify/wiki/Simplify.DI](https://github.com/i4004/Simplify/wiki/Simplify.DI)

A common interface for IOC containers. Decouples users and frameworks (that are based on Simplify.DI) from dependency on IOC containers.

Decouples users and frameworks (that are based on Simplify.DI) from dependency on IOC containers. Instead of that, they will only depend on Simplify.DI interface.

Provides DIContainer.Current ambient context as centralized IOC container. Provides unified Interface for creation IOC container providers (wrappers) for IOC container frameworks.


pleb:
Repo is achieved

PR #817, opened 2019-07-25
Show stars count for all github repositories

SamProf:
CATEGORY/PROJECT_NAME - [PROJECT_LINK](PROJECT_LINK)

PROJECT_DESCRIPTION

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 A collection of awesome .NET libraries, tools, frameworks, and software.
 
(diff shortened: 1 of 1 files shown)


zihotki:
Thank you for the PR, it would be good to have that info but the solution you provided is wrong and the numbers are already wrong. And updating the star counts is going to be a a manual process, that's a no-go.

MaceWindu:
@SamProf , check this https://buttons.github.io/

PR #827, opened 2019-09-03
Add Smaragd

nkristek:
MVVM/Smaragd - [Smaragd](https://github.com/nkristek/Smaragd)

This is a very lightweight library containing base classes for implementing .NET applications using the MVVM architecture.


nkristek:
Thank you for looking into my project. I noticed the tag indicating it may not contain enough functionality. I just want to assure that I included everything that is needed to build a feature rich MVVM application. If it lacks something someone needs in a MVVM framework, I'm more than happy to discuss and/or include it, but I generally kept it down feature wise, since I want to minimize the general footprint (performance reasons and feature creep).

PR #828, opened 2019-09-03
Added Cloudsmith (Package Management SaaS)

lskillen:
Package Management / [Cloudsmith](https://cloudsmith.io)

Hello! This adds [Cloudsmith](https://cloudsmith.io), which is a fully managed package management SaaS, with support for a full range of packaging technologies (but most recently, and importantly, NuGet). It's commercial, but is also completely free for open-source software and for public-only repositories.

Disclosure: I work for Cloudsmith, and I think it is awesome. :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -786,6 +786,7 @@ metadata in media files, including video, audio, and photo formats
 
 * [NuGet](https://www.nuget.org/) - The .NET package manager
 * [BaGet](https://github.com/loic-sharma/BaGet/) - A cross-platform, lightweight NuGet and Symbol server
+* [Cloudsmith](https://cloudsmith.io/l/nuget-feed/) - A fully managed package management SaaS, with support for NuGet, Npm, Docker and much more. **[Free for Public/OSS]** **[$]**
 * [MyGet](https://www.myget.org/) - Hosted Package Repository for NuGet, NPM, Bower and VSIX. Also provides CI as-a-Service. **[$]**
 * [Paket](https://github.com/fsprojects/Paket) - A package dependency manager for .NET with support for NuGet packages and GitHub repositories. https://fsprojects.github.io/Paket/
 * [Sleet](https://github.com/emgarten/sleet/) - A NuGet v3 static feed generator with support for AWS S3 and Azure Storage


Jogai:
Cool that its free, but really geared towards enterprisey stuff. I doubt that it will be awesome in an open source environment. They will pull from public registries like nuget/npm etc. because all the open source packages live there.

lskillen:
@Jogai There's no doubt that Cloudsmith has Enterprise customers, but I think you'd be surprised about how involved it is with open-source. :-) Usually our open-source users are distributors (e.g. https://isc.org, who are powering some part of our conversation right now), and are doing something a bit more custom than what the public registries support (e.g. a custom domain, or they're offering paid packages on top of open-source).

For everyone else, it's rare that an entire stack is completely open-source, and users want somewhere to store their proprietary packages (and to distribute from). In other cases, it's because the public registries are either very controlled (e.g. release cadence is slow), or limited (e.g. don't allow snapshots for Java), or perhaps even unreliable/slow.

So there are use-cases for public and private distributors, plus public and private users. :-)

Jogai:
Thanks for following up.

> it's rare that an entire stack is completely open-source

I was referring to OS projects. They have by definition an entire OS stack.

> extremely generous amount provided for open-source users

And that would be enterprises (or just companies) that incorporate OS in their stack I think.

I see the drawback of the public registries, but I don't see many companies maintain cloudsmith just to overcome just those. And if companies need a lot of cloudsmith's features they probably are bigger.

lskillen:
@Jogai Is the OSS-only argument completely valid here? It's worth bearing in mind that Cloudsmith isn't establishing a new category here; we're just adding to the already existing category of Package Management which services such as BaGet, MyGet and Sleet sit under. MyGet is also commercial. The _reason_ to add Cloudsmith is that it provides a significant (nay, _awesome_) amount of value regardless of whether you pay for it, or you don't.

Cloudsmith sits in the "developer tools" market and targets anything from individual developers through to the Enterprise. Building libraries internally that don't get shipped to nuget.org? You can use Cloudsmith. Deploying libraries/services internally to servers? You can use Cloudsmith. Selling NuGet packages to customers as a vendor, and you need a way to control/distribute that? You can use Cloudsmith. Different use-cases for different people and situations. :-)

Jogai:
True, its not a OS only list.

PR #843, opened 2019-10-22
Fixed Some 301 redirects

fxredeemer:
Fixed some 301 Redirects by replacing with url suggested in travis


fxredeemer:
Closed due to PR #844  Fixed all 503 and 404, that fixes even more issues

PR #846, opened 2019-11-01
Add ScreenToGif

infectormp:
Image Processing/ScreenToGif - [ScreenToGif](https://github.com/NickeManarin/ScreenToGif)

Open source .NET tool allows you to record a selected area of your screen, live feed from your webcam or live drawings from a sketchboard.


Jogai:
It certainly is awesome, but belongs to a list like this: https://github.com/Microsoft/dotnet/blob/master/dotnet-developer-projects.md

PR #864, opened 2020-01-16
Added HTML Minifier tool

MarioZ:
Minification/HTML Minifier - [HTML Minifier](https://github.com/deanhume/html-minifier)

A simple command line tool to minify your HTML, Razor views & Web Forms views. By minifying your HTML on bigger web pages, you will save on bytes that your users need to download. Fewer bytes equal faster web pages & faster web pages equal happy users!


Jogai:
I dont think this is awesome. Its reimplementing html parsing with regex, and I dont know how well tested this all is. Webmarkupmin does the same but seems more solid.

MarioZ:
I agree, the tool is not "awesome", but I think it's useful.
Does "Web Markup Minifier" support CSHTML and ASPX?

Jogai:
Its certainly useful. And no, it doesnt do template languages. It has extensions to minify the output of web forms/mvc and more.

pleb:
Sorry, but the community hasn't rallied behind this project to get it included. Thanks for your contribution.

PR #896, opened 2020-04-08
Added the SqlServer, MySql, SqLite, PostgreSql, SqlServer (Bulk) packages.

mikependon:
[RepoDb](http://repodb.net) - [https://github.com/mikependon/RepoDb](https://github.com/mikependon/RepoDb)

**PROJECT_DESCRIPTION**

A hybrid ORM Library for .NET.

**CHANGES**

The RepoDb is now supporting the MySql, SqLite, PostgreSQL and the SQL Bulk Operations. In addition, as part of the organizations of the code-lines and packages, we have separated the entries of the SQL Server into its dedicated package named RepoDb.SqlServer.

Please allow us to push these updates. Thanks!


pleb:
@mikependon I'm not sure why a link per supported database vendor is required. I would have thought, from the RepoDb page, one would be able to any specific database vender links they need. In my opinion, one link should suffice, but as this a community list I'll take it so that people are welcome to comment and champion the cause. Good luck

mikependon:
@pleb - thank you for this comment. I think you are right, one entry for core would suffice here. By now, I looked at the README and only the main entry is present there, and that is okay. I will close this request.

PR #928, opened 2020-07-28
Add CsCheck

AnthonyLloyd:
Testing/CsCheck - [https://github.com/AnthonyLloyd/CsCheck](https://github.com/AnthonyLloyd/CsCheck)

Random testing library for C#. Including multithreading, performance and regression testing.

Innovative random shrinking means it's fast, runs multithreaded, and is easier to compose Gen in C# as there is no need to for Arbitrary.


AnthonyLloyd:
@pleb Just updated and rebased this PR. CsCheck is now 2.0.0 with concurrency testing with full shrinking and causal profiling. These are not seen anywhere else in .NET. Closest would be [Coyote](https://github.com/microsoft/coyote) which is quite a different beast. Possibly it should be added also but I have little experience of it to say.

PR #994, opened 2021-03-01
Added JSON formatter and validator to tools section

ThomasArdal:
Tools/JSON Formatter and Validator - [https://elmah.io/tools/json-formatter/](https://elmah.io/tools/json-formatter/)

A blazing fast JSON formatter and validator that won't share JSON with a server.


Jogai:
More awesome tools: https://elmah.io/tools/

But I doubt it belongs on a _dotnet_ list..

ThomasArdal:
> But I doubt it belongs on a _dotnet_ list..

Not sure I understand? Do you mean that the JSON formatter doesn't belong on the list or some of the other tools on that page?

Jogai:
The json formatter. Its not specific to .net. Formatting is even build into visual studio.

ThomasArdal:
Sure. So is XML formatting, etc. I often find myself googling tools like this. I don't want to be the judge over if this tool should be added to the list or not, of course.

Jogai:
True. Its just a consideration.

PR #999, opened 2021-03-16
Add PdfSharp

aljones:
PDF/PdfSharp- [http://www.pdfsharp.net](http://www.pdfsharp.net)

Create and process PDF documents. The same drawing routines can be used to create PDF documents, draw on the screen, or send output to any printer.


pleb:
Is this still maintained? The website seems a bit out of date, as I found a reference to codeplex

PR #1003, opened 2021-03-28
Add EBind

SIDOVSKY:
MVVM/EBind - [EBind](https://github.com/SIDOVSKY/EBind)

A data-binding library that uses collection initializer syntax and linq expression trees to set up bindings in a concise way.
EBind interprets expressions to delegates which makes it very fast.
It's highly configurable, extensible, and safe for the mono-linker and AOT compilation.


Jogai:
Looks good to me

PR #1024, opened 2021-08-15
Add Plastic

sang-hyeon:
Application/Plastic - [Plastic](https://github.com/sang-hyeon/Plastic)

Plastic provides encapsulation of things like Domain, Application Rules, Business Rules or Business Logic in Application.
Command pattern is used to do this, and It uses a source generator to provide flexible code.


deaquino:
Hi, 
Just checked the code of Plastic. I think it is in very early stage.

I find it an approach pretty similar to MediatR or so, but in order to use it, I'd like to have more documentation and rich examples.

Sorry for the delay in reply to this thread.

sang-hyeon:
Thanks.

I'm preparing more detailed documentation and planning a big roadmap.  this project has been influenced by MediatR, but the concept is diffrent.

The diffrence will become clear.

PR #1036, opened 2021-10-22
Add online decompiler

cosmolev:
Decompilation/Decompiler.com - https://www.decompiler.com/

Online .NET decompiler.


cosmolev:
> Nice tool. However, I couldn't see any info around who made this, or policies for handling uploaded files.
> 
> I'm reluctant to merge this until more is known.

I made this tool - it's my hobby project. Uploaded files are deleted immediately after the decompilation. Decompilation results are stored for not more than 2 days, but usually deleted earlier than that (exact time depends on the amount of free disk space). But there is a "Delete" button that deletes everything immediately.

Is this info good enough or do you suggest I should also publish it on the website itself?

PR #1059, opened 2022-01-06
Stowage

aloneguid:
Database Drivers/Stowage - [https://github.com/aloneguid/stowage](https://github.com/aloneguid/stowage)

Stowage is a bloat-free .NET cloud storage kit that supports at minimum THE major ☁ providers.


Jogai:
This looks really awesome! One question, does it make storage.net obsolete?

aloneguid:
@Jogai absolutely not, they are built in a very different way. Storage.Net uses official fat SDKs and Stowage is super lightweight.

PR #1074, opened 2022-03-08
Added vvvv - a visual live-programming environment for vvvv

joreg:
IDE/vvvv - [https://visualprogramming.net](https://visualprogramming.net)

vvvv is a visual live-programming environment with its own (visual) language VL and library built on top of the .NET libraries. It can consume any .NET NuGet or .dll directly and makes functions and classes available as visual nodes that can be connected to build programs. The term "live-programming" is used to describe the fact that every change made to a program is immediately applied to the running program using a state hot-reload approach.

EDIT: vvvv is free for non-commercial use without any restrictions. Commercial use requires a license.


microdee:
yeeeah boi! I endorse!

joreg:
@pleb is there a reason this was not accepted?

pleb:
@joreg apologies, I can't remember to be honest.

PR #1088, opened 2022-06-13
Added RPC on with gRPC official GitHub repo

GevorDanielyan:
gRPC/gRPC on .NET - [gRPC on .NET](https://docs.microsoft.com/en-us/aspnet/core/grpc/?view=aspnetcore-6.0)

Overview for gRPC on .NET

gRPC is a language agnostic, high-performance Remote Procedure Call (RPC) framework.


pleb:
Other mods please chime in.

I would suggest a category of RPC rather than the specific gRPC.

I'd also suggest linking the .net library repo and not an MS article.

GevorDanielyan:
Dear @pleb ,

Thanks for your review և suggestion. I recently started using this technology in .NET, I created a trading platform, I would definitely recommend using this technology, so I decided to add this section as well. I can change gRPC to RPC, I will change Microsoft's URL to grpc-dotnet URL if you do not mind.

pleb:
@GevorDanielyan yes, please go ahead with the changes 👍

PR #1092, opened 2022-06-21
Switch from travis to GitHub actions

pleb:
Switches from Travis (which is currently not working) to GitHub actions


pleb:
Link #1092

PR #1105, opened 2022-07-19
Update README.md: Added Akade.IndexedSet

akade:
Algorithms and Data Structures/Akade.IndexedSet - [Akade.IndexedSet](https://github.com/akade/Akade.IndexedSet)

Akade.IndexedSet is a library that provides a convenient data structure supporting efficient in-memory indexing and querying, including range queries and fuzzy string matching.

The library is MIT-licensed, is actively developed by me and distributed via NuGet. While it is early in its developement, I think that it is now mature enough to provide actual value to devs. And to the best of my knowledge, I do not know of a similar library in our C#/.Net ecosystem...

Thanks for your consideration. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ To the extent possible under law, [Vitali Fokin](https://github.com/quozd) has w
 * [Algorithmia](https://github.com/SolutionsDesign/Algorithmia) - Algorithm and data-structure library for .NET 3.5 and up. Algorithmia contains sophisticated algorithms and data-structures like graphs, priority queues, command, undo-redo and more.
 * [C# Algorithms](https://github.com/aalhour/C-Sharp-Algorithms) - A C# plug-and-play class-library project of standard Data Structures and Algorithms. It contains 35+ Data Structures and 30+ Algorithms designed as Object-Oriented isolated components.
 * [Towel](https://github.com/ZacharyPatten/Towel) - data structures, algorithms, mathematics, metadata, extensions, console, measurements, and other useful stuff
+* [Akade.IndexedSet](https://github.com/akade/Akade.IndexedSet) - A convenient data structure supporting efficient in-memory indexing and querying, including range queries and fuzzy string matching.
 
 ## API
 


Jogai:
Is it like lucene with facets?

akade:
Yes and no, the scope is much narrower: It's not supposed to be a fully fledged fulltext search engine that Lucene is - it's supposed to be an easy and convenient way to setup multiple in-memory "indices" without the hassle to maintain the entries for them. So it's a data structure to replace manually setting up & maintaining consistency of for example multiple dictionaries and while providing fuzzy search, it's nowhere near the feautres (and probaby never will) of Lucene or Elastic & Co but much more easy to use. So from the readme:

```csharp
IndexedSet<int, Purchase> set = data.ToIndexedSet(x => x.Id)
                                    .WithIndex(x => x.ProductId)
                                    .WithRangeIndex(x => x.Amount)
                                    .WithRangeIndex(x => x.UnitPrice)
                                    .WithRangeIndex(x => x.Amount * x.UnitPrice)
                                    .WithIndex(x => (x.ProductId, x.UnitPrice))
                                    .Build();

// efficient queries on configured indices
_ = set.Where(x => x.ProductId, 4);
_ = set.Range(x => x.Amount, 1, 3, inclusiveStart: true, inclusiveEnd: true);
_ = set.GreaterThanOrEqual(x => x.UnitPrice, 10);
_ = set.MaxBy(x => x.Amount * x.UnitPrice);
_ = set.Where(x => (x.ProductId, x.UnitPrice), (4, 10));
```
So it can be used on the fly to support certain in-memory calculations but also used to keep data over a longer amount of time, such as in a cache...

Jogai:
I read the examples and tests but its still not entirely clear to me. The example queries could benefit from a comment with the results. I guess the first one returns the product with id 4. But linq could do that just as easily, and same goes for a lot of other examples.

akade:
The difference is that linq will enumerate through all the entries while those queries do not - I'm currently working on benchmarks to illustrate that. But thanks for pointing out that it's not clear enough at the moment: I'll try to update the readme / sample tests to highlight that.

PR #1122, opened 2022-10-17
Update README.md

romanov:
Added to [MISC]

CATEGORY/PROJECT_NAME - [Percentage.NET](https://github.com/romanov/PercentageNET)

PROJECT_DESCRIPTION

Simple library to count percentages in your projects.


pleb:
Seems like a cool library, @romanov. I can't add it to the list because it doesn't have tests and doesn't seem popular within the dotnet community. If both of these change, feel free to submit a new PR. 

Good luck.

PR #1126, opened 2022-10-23
Add Emulators category.

p0008874:
Emulation need more love.


pleb:
Some of the links are to read-only repos

PR #1128, opened 2022-10-25
Database/RocksDB

theolivenbaum:
Database/RocksDB - [RocksDB](https://github.com/curiosity-ai/rocksdb-sharp)

PROJECT_DESCRIPTION
RocksDB is a key-value database with a log-structured-merge design, optimized for flash and RAM storage, which can be tuned to balance write-, read-, and space-amplification factors. This project provides C# bindings to the database, plus native builds for Windows, Linux and macOS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,6 +393,7 @@ To the extent possible under law, [Vitali Fokin](https://github.com/quozd) has w
 
 ## Database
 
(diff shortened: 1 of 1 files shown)


pleb:
Would this be better placed under ## Database Drivers ?

theolivenbaum:
It's more an embedded database than a driver, as it won't talk via network to RockdDB, thus I thought the database section would be a better fit.

PR #1137, opened 2023-01-13
Added FluentSerializer to the readme

Marvin-Brouwer:
Serializing - [FluentSerializer](https://github.com/Marvin-Brouwer/FluentSerializer#readme) - A profile-based Serializer for multiple data-formats.

### PROJECT_DESCRIPTION

`FluentSerializer` is a library to help you with serializing to-and-from C# POCOs using profiles.
The use of profiles helps making it easier to understand how the data vs the code looks in a single glance.
So instead of needing a library like `Mappster` or `AutoMapper` to mold data into your code structure, you can now map that same data into a clean data representation and use the mapping frameworks for what their intended, to translate data.

Next to a clear overview of how the data looks, this library provides you with serializing methods for multiple data formats with a similar api.
So when you're required to tie XML and JSON together at least the code for serializing looks similar across your solution.


pleb:
@Jogai thoughts?

Jogai:
Yep, seems awesome, a good readme, with documentation and makes use of tests.

PR #1220, opened 2024-06-10
Move 4 libraries into new Cloud Storage section

robinrodricks:
I have introduced a new category called "Cloud Storage".

I have moved 4 cloud storage libraries into a this category, as they are more appropriate in that section. They were scattered all over the place (under Database, Distributed Computing, etc).

No libraries have been added or removed.

I think you might find it useful.


robinrodricks:
Thanks a lot!

PR #1252, opened 2024-11-01
Added xsl-transformer CLI tool

armanossiloko:
CLI/xsl-transformer - [https://github.com/armanossiloko/xsl-transformer](https://github.com/armanossiloko/xsl-transformer)

`xsl-transformer` is a simple CLI tool which allows transforming XML data into readable output content such as HTML using the XSL transform technique.

e.g A notable example (and the root cause of its creation) are IIS logs in XML format which used to be readable by simply opening them in Internet Explorer. That, however, has become harder with IE's retirement and opening any of those logs would result in viewing raw XML data, rather than an HTML formatted document.


armanossiloko:
Closing as there's been lack of comments and reviews.

PR #1294, opened 2025-03-25
Remove unmaintained libraries and projects

alastairlundy:
This PR removes unmaintained, discontinued, or otherwise unsupported projects and libraries from the list.

In deciding what constitutes an unmaintained library I've tried to keep with the spirit of this project's contribution guidelines.

For judging whether a library/project is unmaintained I've used several criteria:

- Projects without any commits for several years and where the author has not responded to issues or PRs from within the past few years
- Projects where the author has explicitly discontinued them or explicitly no longer provides support for them
- Projects that are no longer available (E.g. libraries that were on codeplex that now redirect to Microsoft's website, and repos with dead links)
- Projects that have been archived by their author

I've also corrected links to some projects that are maintained but use a new domain or url for their library as well as some projects that have changed their name but are substantively the same project as before.


newbe36524:
I think it would be better to make them unmaintained rather than removing them. 
They all contributed by maintainers. We need to do some respect for their efforts

alastairlundy:
I respectfully disagree. 

At one point in time the pieces of software merited being on the list but being unmaintained there is a danger that somebody will find and use the project despite being unmaintained. It may even prevent people from starting new projects in certain categories if the list stays as it is now.

In addition, unmaintained items are not accepted to be added to the list so why should existing unmaintained items be allowed to remain there if they pose the same risks and dangers and general inconvenience to users of the list.

Their efforts were respected by being on the list whilst they maintained the software. If there is a need to explicitly state respect for their work then a statement could be added to the readme to that effect whilst explaining that unmaintained items aren't on the list. 

E.g. We appreciate the time and effort maintainers have put into maintaining software, and would like to thank the maintainers that previously maintained software on this list.

pleb:
Thanks. Love the clean up

PR #1337, opened 2025-09-08
Add RCParsing into Parser Library section

RomeCore:
Parser Library/RCParsing - [RCParsing](https://github.com/RomeCore/RCParsing)

This is the parser framework that focuses on developer experience first, providing a huge toolkit for making own DSLs of any complexity or even text scrappers. It highlights with these features:

- Ability to parse indent-based languages like Python and YAML with barrier tokens.
- Mode for finding all matches in the text, just like the regex.
- Performance that comparable with Pidgin, which known as one of fastest parser-combinators in the .NET.
- Big set of useful parsing primitives: numbers with configurable flags, escaped text with auto-unescaping, separated lists with ability to allow trailing separator, and more...
- Very detailed documentation with examples and 99% of XML docstrings coverage.

It has been tested on 80+ of test methods and has been already used in one of my projects. It actively maintained.


RomeCore:
And when this PR will be approved? Project has been grown a lot since i submitted it here, it can parse Python 3.13 now! Also, if you will merge it, i want to change title to "The fluent, lightweight and powerful .NET lexerless parsing framework with lots of capabilities"

PR #1454, opened 2026-06-26
Add AutoMap.Generator — compile-time object mapping via Roslyn source generators

Swevo:
## What

Adds **[AutoMap.Generator](https://github.com/Swevo/AutoMap.Generator)** to the **Object to object mapping** section.

## About the library

AutoMap.Generator is a Roslyn source generator that produces strongly-typed extension methods at build time from a single attribute:

\\\csharp
[Map(typeof(OrderDto))]
public class Order { ... }

// Generated:
var dto = order.ToOrderDto();
\\\

- **Zero reflection** — generated code is plain C# property assignment
- **AOT-safe / MAUI-compatible** — no runtime code emission
- **Build-time diagnostics** (AM001–AM007) catch mistakes at compile time
- **netstandard2.0** target — works on .NET 6/7/8/9 and MAUI
- **1k+ NuGet downloads**, MIT licence, active development

## Why it belongs here

The existing section lists AutoMapper (reflection-based) and Mapperly (source generator). AutoMap.Generator is a second source-generator option with a different ergonomic approach (attribute on the domain class, no mapper class needed), making it a genuine alternative that .NET developers searching this list would want to know about.


Swevo:
Hi @quozd — just checking in on this PR. Happy to make any changes needed for it to merge. Thanks for maintaining this list!

PR #1473, opened 2026-07-14
Add StrongTypes to Misc section

KaliCZ:
Adding **StrongTypes** to the Misc section.

**Project:** https://github.com/KaliCZ/StrongTypes
**NuGet:** https://www.nuget.org/packages/Kalicz.StrongTypes/
**Write-up:** https://www.kalandra.tech/blog/zero-code-validations-in-your-dotnet-api/

### What it is

StrongTypes moves validation out of runtime checks and into the type system. Instead of passing a `string` around and re-checking it for emptiness at every layer, you pass a `NonEmptyString` and the signature carries the guarantee. It includes:

* Validated scalars: `NonEmptyString`, `Email`, and `Positive<T>` / `NonNegative<T>` / `Negative<T>` / `NonPositive<T>` over any `INumber<T>`
* `NonEmptyEnumerable<T>`, a sequence guaranteed non-empty, with LINQ operators that preserve the invariant and total `Max` / `Min` / `Last` overloads that can't throw on empty input
* Interval types (`FiniteInterval<T>`, `IntervalFrom<T>`, `Interval<T>`, ...) with configurable bound inclusivity, containment and overlap
* `Maybe<T>` and `Result<T, TError>` for what nullable reference types can't express, such as HTTP PATCH's three-way "not supplied" vs "explicitly null" vs "new value"

Every type has a `System.Text.Json` converter attached via `[JsonConverter]`, so invalid input fails at deserialization with no registration and no custom `JsonSerializerOptions`. Companion packages cover EF Core persistence, ASP.NET Core model binding, OpenAPI schema generation (both `Microsoft.AspNetCore.OpenApi` and Swashbuckle), FsCheck generators, and WPF/MVVM binding.

[This write-up](https://www.kalandra.tech/blog/zero-code-validations-in-your-dotnet-api/) walks the whole before/after on one order endpoint: the same rule written three times across DTO, controller and service, then the version where the request record takes `Email` / `NonEmptyString` / `Positive<int>`, the guard clauses are simply deleted, and the OpenAPI schema still emits `format: email`, `minLength` and `exclusiveMinimum` for free.

### Why I think it fits

Misc already carries FluentValidation and Validot, which attack the same problem from the rule-based side: validate a model against a rule set at a boundary. StrongTypes is the complementary approach, sometimes called "parse, don't validate": encode the invariant in the type, check it once at construction, and let the compiler carry it from there. Readers browsing that part of the list seem like the right audience, which is why I've slotted it next to them.

### Against the quality standards

* **Actively maintained.** v1.2.0 released this week, regular cadence since 1.0, CI green on every push.
* **Stable.** 1.0 since April 2026, semver, 7 packages published to NuGet.
* **Documented.** Full readme with per-feature examples and diagrams, a readme per companion package, plus the write-up above.
* **Tests.** Around 1,300 test methods, including integration tests that round-trip every type through the real ASP.NET Core pipeline and EF Core against both SQL Server and PostgreSQL via Testcontainers, plus assertions on the generated OpenAPI documents.

### Disclosure

I'm the author. It's MIT-licensed and fully open source with no commercial tier, so none of the `[$]` / `[Proprietary]` / `[Free for OSS]` tags apply.

It's also a young project (first release April 2026, ~2.4k downloads on the core package), so I understand if you'd rather it build more traction first. Happy to reword the entry or move it elsewhere.


KaliCZ:
@quozd @pleb Can I kindly ask you to look at this one?

